PR #16, opened 2013-10-20
Fixing build

helloIAmPau:
Hi Alan,
I'm sorry for the delay but I've a lot of work in this period.

I've fixed the issues you have previously open regarding the compilation. Now the makefile should build the deps using your "method".

If it's ok, I begin to import some tests. 

PS I've also created a Vagrant box to use your add-on on others OS. Let me know if you are interested on it.


binding.gyp
```diff
@@ -151,7 +151,8 @@
               '../<(libwebrtc_out)/../libvideo_render_module.a',
               '../<(libwebrtc_out)/../libvoice_engine.a',
               '../<(libwebrtc_out)/../libwebrtc_opus.a',
-              '../<(libwebrtc_out)/../libwebrtc_utility.a'
+              '../<(libwebrtc_out)/../libwebrtc_utility.a',
+              '-lssl'
```
modeswitch, line 155: Does the system openssl need to be linked here? We should be linking against the version in libjingle `'../<(libwebrtc_out)/third_party/openssl/libopenssl.a'`.

helloIAmPau:
In Mac libjingle uses system openssl because it is present out of the box.

Issue #26, opened 2013-10-27
replace deprecated V8 APIs


devnullmail:
compiling node-webrtc against the current joyent/node repository fails (e.g. 2013-10-26).

node-webrtc uses V8 APIs that are deprecated. These have been removed in V8 3.21+.

```
gyp info spawn args [ 'V=1', 'BUILDTYPE=Release', '-C', 'build' ]
  g++ '-D_LARGEFILE_SOURCE' '-D_FILE_OFFSET_BITS=64' '-DLARGEFILE_SOURCE' '-DWEBRTC_TARGET_PC' '-DWEBRTC_LINUX' '-DWEBRTC_THREAD_RR' '-DEXPAT_RELATIVE_PATH' '-DGTEST_RELATIVE_PATH' '-DJSONCPP_RELATIVE_PATH' '-DWEBRTC_RELATIVE_PATH' '-DPOSIX' '-D__STDC_FORMAT_MACROS' '-DDYNAMIC_ANNOTATIONS_ENABLED=0' '-DBUILDING_NODE_EXTENSION' -I/home/devnullmail/node/src -I/home/devnullmail/node/deps/uv/include -I/home/devnullmail/node/deps/v8/include -I../lib/libwebrtc/trunk -I../lib/libwebrtc/trunk/third_party/webrtc -I../lib/libwebrtc/trunk/third_party/webrtc/system_wrappers/interface -I../lib/libwebrtc/trunk/third_party  -fPIC -pthread -Wall -Wextra -Wno-unused-parameter -m64 -pthread -fno-exceptions -fno-strict-aliasing -Wall -Wno-unused-parameter -Wno-missing-field-initializers -Wextra -Wno-unused-local-typedefs -Wno-uninitialized -Wno-unused-variable -Wno-unused-but-set-variable -pipe -fno-ident -fdata-sections -ffunction-sections -fPIC -fpermissive -O3 -ffunction-sections -fdata-sections -fno-tree-vrp -fno-rtti -fno-exceptions -MMD -MF ./Release/.deps/Release/obj.target/webrtc/src/binding.o.d.raw   -c -o Release/obj.target/webrtc/src/binding.o ../src/binding.cc
In file included from ../src/binding.cc:4:0:
../src/peerconnection.h:176:35: error: ‘Arguments’ does not name a type
   static Handle<Value> New( const Arguments& args );
                                   ^
[...]

Issue #17, opened 2013-10-20
Consider using NAN


DamonOehlman:
Not sure if this will be of interest, but I know @rvagg built it to make writing native bindings to things more portable across node versions:

https://github.com/rvagg/nan

Additionally, it's worth noting that this was created to assist with the work Rod does with [leveldown](https://github.com/rvagg/node-leveldown) so it's been built and tested around a solid use case.

Just something to consider as I know the breadth of the underlying webrtc c++ library is pretty massive so anything that means writing less code is probably a good thing.

rvagg:
Actually, it's **[ws](https://github.com/einaros/ws)** that gives NAN its hardest workout. Pretty much anyone that wants to use WebSockets with Node ends up installing ws. **[canvas](https://github.com/LearnBoost/node-canvas)** probably comes in second. **[zeromq](https://github.com/JustinTulloss/zeromq.node/pull/221)** is just waiting for Travis to get their act into gear with 0.11 before they merge in NAN to manage forward version compatibility.

modeswitch:
Yeah, that seems reasonable. Not super high-priority at the moment, but a good step for improving portability.

DamonOehlman:
Sure, I agree.  Just thought I'd point it out and let you know that Rod is super helpful if and when you do go down that path :)

modeswitch:
Merged a pull from @kkoopa that fixes this.

DamonOehlman:
Sweet - that's awesome :)

PR #27, opened 2013-10-30
Ported to NAN, compiles on 0.11.8.

kkoopa:
In response to #17 and #26, here is a patch that makes use of NAN.

### DIFF
--- a/src/datachannel.cc
+++ b/src/datachannel.cc
@@ -70,7 +70,7 @@ void DataChannel::Run(uv_async_t* handle, int status)
   TRACE_CALL;
   NanScope();
   DataChannel* self = static_cast<DataChannel*>(handle->data);
-  v8::Local<v8::Object> ldc = NanObjectWrapHandle(self);
+  v8::Local<v8::Object> dc = NanObjectWrapHandle(self);
 
   while(true)
   {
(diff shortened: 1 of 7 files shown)


modeswitch:
This looks good. Does the `bridge.js`/`peer.html` test pass?

kkoopa:
Unfortunately, I have no idea. Did the porting in a VM without X or many other things, so couldn't run Chrome, and I've never used webrtc, so I don't have any testing-code lying around. The module loaded and printed a couple lines when running tests/test.js. If someone with more insight and opportunity could do a test it would be great. I'll fix any problems that were introduced in the process, but usually things work if everything compiled successfully.

modeswitch:
This isn't building on my machine:

```
make: Entering directory '/home/ack/work/node-webrtc/build'
  CXX(target) Release/obj.target/webrtc/src/binding.o
  CXX(target) Release/obj.target/webrtc/src/peerconnection.o
In file included from ../src/peerconnection.h:19:0,
                 from ../src/peerconnection.cc:16:
../src/peerconnection.cc: In static member function ‘static void PeerConnection::Run(uv_async_t*, int)’:
../node_modules/nan/nan.h:394:40: error: conversion from ‘v8::Persistent<v8::Object>’ to non-scalar type ‘v8::Local<v8::Object>’ requested
 # define NanObjectWrapHandle(obj) obj->handle_
                                        ^
../src/peerconnection.cc:146:31: note: in expansion of macro ‘NanObjectWrapHandle’
   v8::Local<v8::Object> lpc = NanObjectWrapHandle(self);
                               ^
webrtc.target.mk:147: recipe for target 'Release/obj.target/webrtc/src/peerconnection.o' failed
make: *** [Release/obj.target/webrtc/src/peerconnection.o] Error 1
make: Leaving directory '/home/ack/work/node-webrtc/build'
```

Maybe I'm missing something?

src/peerconnection.cc
```diff
@@ -140,10 +140,10 @@ void PeerConnection::QueueEvent(AsyncEventType type, void* data)
 void PeerConnection::Run(uv_async_t* handle, int status)
 {
   TRACE_CALL;
-  HandleScope scope;
+  NanScope();
 
   PeerConnection* self = static_cast<PeerConnection*>(handle->data);
-  v8::Persistent<v8::Object> pc = self->handle_;
+  v8::Local<v8::Object> lpc = NanObjectWrapHandle(self);
```
modeswitch: Why was this renamed?

kkoopa: I thought the persistent handle would be needed as well as the local, but it was not referenced anywhere, so I removed it and neglected to rename back. I've fixed that now.

kkoopa:
What version of Node are you using? Nevermind, it is 0.10.something. It seems that the problem is that the async data is a `Persistent<Object>` under 0.10, but a `Local<Object>` under 0.11.

0.10:
`v8::Local<v8::Object> pc = NanPersistentToLocal(NanObjectWrapHandle(self));`

0.11:
`v8::Local<v8::Object> pc = NanObjectWrapHandle(self);`

cc @rvagg: What's your take on this.

modeswitch:
I'm using v0.10.21. I'll have a look this afternoon.

src/peerconnection.cc
modeswitch: Cool, if you push to your branch I'll merge to incoming.

modeswitch: Also, I think this line should be `v8::Persistent<v8::Object> pc = NanObjectWrapHandle(self);`

src/datachannel.cc
```diff
@@ -69,9 +68,9 @@ void DataChannel::QueueEvent(AsyncEventType type, void* data)
 void DataChannel::Run(uv_async_t* handle, int status)
 {
   TRACE_CALL;
-  HandleScope scope;
+  NanScope();
   DataChannel* self = static_cast<DataChannel*>(handle->data);
-  v8::Persistent<v8::Object> dc = self->handle_;
+  v8::Local<v8::Object> ldc = NanObjectWrapHandle(self);
```
modeswitch: Should be `v8::Persistent<v8::Object> dc = NanObjectWrapHandle(self);`.

modeswitch:
Also, please add yourself to the AUTHORS file!

modeswitch:
Tests run with those changes, so once you update your pull request I'll merge.

src/peerconnection.cc
kkoopa: It is these Persistent assignments that are problematic. You cannot assign to a persistent in new Node, so that will not work for 0.11.4+. The problem is that the wrapped object is persistent under 0.10 but not under 0.11, so the branches diverge. I have not seen this behavior before and am wondering why it manifests.

kkoopa:
I think I've fixed it. Now it compiles on both 0.10 and 0.11. Have a go and see if it works for you too.

Issue #51, opened 2014-01-22
Updated examples using `dist/wrtc.js` are broken


modeswitch:
```
Uncaught ReferenceError: mozRTCIceCandidate is not defined wrtc.js:2
Uncaught ReferenceError: require is not defined test.js:1
```

modeswitch:
@piranna Can you add some code to build.js to generate dist/wrtc.js using browserify?

piranna:
Yes, I can, it's easy. I didn't do it and just uploaded a generated file
since it's almost impossible it will change ever :-P

2014/1/23 Alan K notifications@github.com

> @piranna https://github.com/piranna Can you add some code to build.js
> to generate dist/wrtc.js using browserify?
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/js-platform/node-webrtc/issues/51#issuecomment-33127683
> .

## 

"Si quieres viajar alrededor del mundo y ser invitado a hablar en un monton
de sitios diferentes, simplemente escribe un sistema operativo Unix."
– Linus Tordvals, creador del sistema operativo Linux

modeswitch:
I believe this is fixed now.

PR #53, opened 2014-01-22
Fix streamtest by updating wrtc.js and adding /dist to static paths.

JessieAMorris:
This is for #51. This fixes all the examples that use the wrtc.js file.


piranna:
dist/wrtc.js was autogenerated with browserify using as basis lib/browser.js, changes should be done instead. By the way: that `global` var, where does it come from? Why don't use window instead?

modeswitch:
@piranna examples work now, but could you fix the autogenerated file? i'm not sure how it was generated.

JessieAMorris:
@piranna The `global` var comes from `this` inside of a function that has no this bound to it. By default, `this` inside of a function defaults to the global variable, whatever that happens to be. In browser situations, that happens to be window. Using the `global` var this way allows us to use the library inside of WebWorkers as well, where `window` is undefined.

piranna:
So global is equivalent to this, window and self (in that context)? I never
seen it before, is it implemented by browsers like self or it's a "manual"
addon?

2014/1/23 Jessie A. Morris notifications@github.com

> @piranna https://github.com/piranna The global var comes from thisinside of a function that has no this bound to it. By default,
> this inside of a function defaults to the global variable, whatever that
> happens to be. In browser situations, that happens to be window. Using the
> global var this way allows us to use the library inside of WebWorkers as
> well, where window is undefined.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/js-platform/node-webrtc/pull/53#issuecomment-33140165
> .

## 

"Si quieres viajar alrededor del mundo y ser invitado a hablar en un monton
de sitios diferentes, simplemente escribe un sistema operativo Unix."
– Linus Tordvals, creador del sistema operativo Linux

JessieAMorris:
It's a "manual" addon defined on line 2 of wrtc.js. But yes, in this case, `global = this = self = window;`.

PR #63, opened 2014-02-01
Tape tests

DamonOehlman:
Hey Alan,

OK - this is a bit of a monster in terms of commit count, but it should merge cleanly against the `develop` branch as I've been ensuring this branch is always up-to-date with changes there.

Essentially - this brings in the start of a suite of tests that should allow us to bring in CI into the mix - which will be awesome!  As per other node modules the tests can be invoked by running `npm test`.

If you have any questions, let me know :)

Cheers,
Damon.


modeswitch:
Could you rebase and pull again?

DamonOehlman:
Good call Alan.  I'll either do that or just create a new branch from develop and add the tests.  Either one of those would be much cleaner.  Will close this PR and open another one in a few hours when I get to coding again.

Issue #67, opened 2014-02-09
Tape tests help


wouldgo:
Hi again,

I forked this repository yesterday to rewrite the building/cleaning scripts and to introduce grunt in the development (as reported in issue #56).

The building script seems good but, when I introduced tape in the default task, it complains for 5/6 failures (I don't understand why sometimes 5 and sometimes 6).
I can't really understand where they came out from (probably I'm too fool for tape testing suite).

Since my aim is a pull request for these functionalities, but some tests are failed, could you please help me to understand the meaning of this output?
![grunt-with-tape](https://f.cloud.github.com/assets/1098882/2120675/936ba160-91bf-11e3-8db9-97870fd2f558.png)
This is the [Gruntfile.js](https://github.com/wouldgo/node-webrtc/blob/develop/Gruntfile.js) (the funny thing is that grunt wrote down _"Done, without erorrs"_  :cake: ).

Thanks,
Dario.

piranna:
You can make the pull-request and update it later, this way we can help you with your progress. Unluckily gclient sync always fails to me, so I couldn't be able to check it :-/ If you have random error they could be due to race conditions due to async operations. Regarding to grunt, it says everything is alright because you didn't told him there was an error, put anything as parameter of the grunt.done() function on the tape testing task.

modeswitch:
@wouldgo Yeah, please make a pull request and I can help you out in there.

@piranna What sync errors are you still having?

piranna:
@modeswitch from a git update copy and after removing the third_party folder this is what I get on the console on several consecutives executions of `npm install`:

```
[piranna@Latitude:~/Dropbox/Proyectos/node-webrtc]
 > npm install

> wrtc@0.0.9 install /home/piranna/Dropbox/Proyectos/node-webrtc
> node bin/build.js

depot_tools
.depot_tools not required
gclient_config
.
gclient_config complete
gclient_sync
...............................................................................................................................................................................................................
gclient_sync failed: 1 null

[piranna@Latitude:~/Dropbox/Proyectos/node-webrtc]
 > npm install

> wrtc@0.0.9 install /home/piranna/Dropbox/Proyectos/node-webrtc
> node bin/build.js

depot_tools
.depot_tools not required
gclient_config
.
gclient_config complete
gclient_sync
.................................................................................
gclient_sync failed: 1 null

[piranna@Latitude:~/Dropbox/Proyectos/node-webrtc]
 > npm install

[...]

Issue #56, opened 2014-01-23
Use Grunt?


piranna:
If we start to add tasks like browserify to the toolchain as talked on issue #51, maybe we could start to use a task manager like Grunt or similar instead of custom-made scripts.

modeswitch:
Agreed.

wouldgo:
Can I now verify all the .js files with jshint again, linting them?

modeswitch:
Yep.

modeswitch:
Fixed.

piranna:
I think this should be re-open and not closed until both bin/clean.sh and bin/build.sh are removed and integrated inside Gruntfile.js and/or build.gyp

wouldgo:
I agree with @DamonOehlman as reported [here](https://github.com/js-platform/node-webrtc/pull/68#issuecomment-34717129): the aim of grunt is on developer part, those scripts are used for installing the module.

Bye.

piranna:
My proposal to use grunt was to recude failure surface. I agree with that comment, but current build is dificult, so I still think we should use grunt everywhere. Another thing is if we have pre-compiled binaries easily installable with a simple script, so not-usual builds can be done changuing to "developer mode".

Issue #42, opened 2013-12-31
Offer official windows support.


kofd:
As long as this project remains in the scope of a relay server or a data channel this may not be necessary, but if this api exposes a mechanism for streaming video from a remote server there are definite use cases for offering the ability to capture and stream applications from a windows instance.

modeswitch:
My intent has always been to support Windows, but for the moment focus is on Linux in the early stage. I agree, there are clear use-cases for Windows.

guymguym:
:+1:

modeswitch:
I don't have access to a Windows machine, so a contributor who can work on this is needed.

kanecko:
![](https://assets-cdn.github.com/images/icons/emoji/unicode/1f44d.png)

feross:
Just wanted to let ya'll know that [WebTorrent](https://github.com/feross/webtorrent) is depending on `wrtc` so that node users can seed torrents to web peers over WebRTC. So, there is a use case for using the data channel on windows – p2p protocols based on webrtc that have non-browser clients.

Another awesome project using `wrtc` (through `simple-peer`) is [webcat](https://github.com/mafintosh/webcat) by @mafintosh. Windows support makes sense here too :-)

rosskukulinski:
@feross neither @modeswitch or I have easy access to a Windows machine -- do you or @mafintosh have access to Windows to help with this?

modeswitch:
I would love to add Windows support. The major effort is updating the libwebrtc build process (which might end up just working) and changing the build script in this project to handle it.

As @rosskukulinski mentioned, I don't have a Windows machine and Travis doesn't offer automated tests for Windows either. I'm happy to mentor anyone who wants to take this on though.

guymguym:
you can easily setup a windows VM on your Mac using virtualbox - windows images were published freely for the sake of IE multiplatform testing - https://www.modern.ie/en-us/virtualization-tools

No9:
@modeswitch OK just ran into this at work
I have had a working chrome build on windows so this should be possible and I should be able to put some "official" work hours into it next week :)

alaa-eddine:
+1 for windows support here :)
I just added webRTC support to my RPC library (https://github.com/Ezelia/eureca.io) which is very useful for people who want to implement multiplayer servers (as WebRTC can use UDP). for now webRTC is only supported on linux.

I have an old windows box too, I started working on a windows support but still having issues to get a working chromium build.

[...]

PR #68, opened 2014-02-10
Building scripting + grunt integration

wouldgo:
Little summary:
- As reported in #67 I refactored the building/cleaning scripts (I didn't receive any error now).
- I introduced grunt, with two task for this first version: jshint and tape runners.

Waiting for feedbacks for pull request eligibility.

Greetings, 
Dario.


piranna:
Good job :-) Just to note: since we are now using grunt, we should use grunt-gyp and grunt-contrib-clean plugins.

wouldgo:
No problem using these tasks.
As soon as I complete this week iteration at work I'll able to introduce and configure :)

piranna:
Cool then :-)

modeswitch:
There's some overlap here with #69.

wouldgo:
From a brief review (I'm on train returning home) the differences between mine and his are mainly the fact I use make as compilation tool (since I found that sometimes ninja fails due a python problem in import) and I use some options for gclient unused in the source build.js.

Waiting some feedbacks considerations.

DamonOehlman:
Hey All,

I guess the approach I was taking in #69 was to look at trying to get us some way down the path towards being able to have the module build on Windows (see #42).  From the perspective of running gclient and bootstrapping the build it definitely seems to work on Windows - I gave it a run on  a Windows surface tablet last night.  There is still a little way to go but I can successfully compile all the libwebrtc dependencies so that's a good start.

From a sense of reviewing this PR, there seems to be a lot in it so it's difficult to look at the install build script in isolation.  So my suggestion would be to that if @modeswitch is happy with the general direction and content of this PR to merge it and I'll revisit #69 again down the track if it building on windows becomes a higher priority.  I'm not on Windows so I'm happy to wait :)

Cheers,
Damon.

modeswitch:
There's a lot happening in this patch. It would be helpful to split it out into smaller pieces.

@DamonOehlman I've put windows support in Milestone 0.2, so I'm not rushed to get that in.

DamonOehlman:
@modeswitch Cool - I think that's wise given what I had to go through last night.  I followed the [Building Chromium with Visual Studio Express](http://www.chromium.org/developers/how-tos/build-instructions-windows#TOC-Setting-up-the-environment-for-building-with-Visual-C-2010-Express-or-Windows-7.1-SDK) instructions and I think I probably downloaded half the Internet getting all the dependencies...

wouldgo:
@modeswitch except the build.js+clean.js and grunt there isn't so much in this patch (all the other modifications are due jshint complains).

Anyway I'm open to review my work.

modeswitch:
@wouldgo Could you split the jshint work into a separate pull?

wouldgo:
done.

binding.gyp
```diff
@@ -19,7 +19,7 @@
       'target_name': 'webrtc',
       'dependencies': [],
       'variables': {
-        'libwebrtc_out%': '<(libwebrtc)/out/Release/obj',
+        'libwebrtc_out%': '<(libwebrtc)/out/Release/obj.target',
```
modeswitch, line 22: What's the purpose of this?

package.json
```diff
@@ -19,11 +23,16 @@
...
+    "tape": "~2.4.2",
     "ws": "~0.4.31",
-    "tape": "~2.4.2"
+    "grunt": "~0.4.2",
+    "grunt-contrib-jshint": "~0.6.*",
+    "grunt-contrib-uglify": "~0.2.4",
+    "grunt-tape": "~0.0.2"
   },
   "scripts": {
     "install": "node bin/build.js",
```
modeswitch, line 38: Since we're now using Grunt, can you convert the build and clean scripts into grunt tasks?

DamonOehlman:
I'd probably advise against using grunt for the actual build and clean scripts themselves.  I think grunt is an excellent tool for helping workflow build tasks that should be undertaken by a maintainer or contributors, but it's not something that people installing the module should have to install as a "non dev" dependency.  Especially considering that for the build to run you'd need to include the `grunt-cli` package also in dependencies.

I realize the original intention of #56 was to probably use grunt in this way, but I guess I'd suggest taking a moment to decide whether that's really what you want.  If it is - then all good :)

wouldgo:
So:
- I've modified the bindings.gyp since make creates an out.target folder instead of out;
- I agree with @DamonOehlman about the grunt considerations. I see that modules as an helper for development tasks (like testing, linting and rebuilding c++ bindings).

PR #69, opened 2014-02-10
Alternative build implementation progress

DamonOehlman:
This is intended as more of a discussion than a straight up pull-request. I know the build process has been discussed in issues so I had a bit of a play around with how I thought I might have tackled the build process, and in particular considering what it might look like as we work towards windows support (#42).  I'm actually giving it a test run on a windows box at the moment, and it's ticking along nicely (so far).

I have taken quite a different approach to what has been done in previous build scripts (this one is not hiding any output from the depot_tools, i.e. gclient).  Additionally I created a simple [node modules](https://github.com/DamonOehlman/node-depot_tools) that is responsible for downloading depot_tools in what I think is a fairly cross platform way (tested on windows).

The actual build script is a bit tighter too, but it might be considered a little bit terse by some (which I understand).  End result of the build script (without touching the node-gyp process) looks something like:

``` js
var async = require('async');
var path = require('path');
var os = require('os');
var fs = require('fs');
var sys = require('sys');
var mkdirp = require('mkdirp');
var exec = require('child_process').exec;
var spawn = require('child_process').spawn;
var gclient = require('depot_tools/gclient');
var ninja = require('depot_tools/ninja');
var PROJECT_DIR = path.resolve(__dirname, '..');
var LIB_WEBRTC_DIR = path.join(PROJECT_DIR, 'third_party', 'libwebrtc');

var tasks = [
  ['config', 'http://webrtc.googlecode.com/svn/trunk'],
  ['sync'],
  ['runhooks']
].map(gclient(LIB_WEBRTC_DIR)).concat([
  ninja(LIB_WEBRTC_DIR, ['-C', 'trunk/out/Release']),
  nodegyp_build
]);

mkdirp(LIB_WEBRTC_DIR, function(err) {
  async.series(tasks, function(err) {
    console.log(err ? 'Build errored: ' + err : 'Build complete');
  });
});

function read_stream(stream)
{
  stream.read();
  process.stdout.write('.');
}

function nodegyp_build(callback)
{
  console.info("nodegyp_build");
  process.stdout.write('.');

  process.chdir(PROJECT_DIR);

  var child = spawn("node-gyp", ["rebuild"]);
  child.stdout.on('readable', read_stream.bind(undefined, child.stdout));
  child.on('exit', function(code, signal)
  {
    process.stdout.write('\n');
    if(0 === code && !signal)
    {
      console.log('nodegyp_build complete');
      callback();
    }
    else
    {
      console.error('nodegyp_build failed:', code, signal);
    }
  });
}
```

### DIFF
--- a/bin/build.js
+++ b/bin/build.js
@@ -1,182 +1,37 @@
 #!/usr/bin/sh
 
+var async = require('async');
 var path = require('path');
 var os = require('os');
 var fs = require('fs');
 var sys = require('sys');
+var mkdirp = require('mkdirp');
 var exec = require('child_process').exec;
 var spawn = require('child_process').spawn;
+var gclient = require('depot_tools/gclient');
+var ninja = require('depot_tools/ninja');
+var PROJECT_DIR = path.resolve(__dirname, '..');
+var LIB_WEBRTC_DIR = path.join(PROJECT_DIR, 'third_party', 'libwebrtc');
 
+var tasks = [
+  ['config', 'http://webrtc.googlecode.com/svn/trunk'],
+  ['sync'],
+  ['runhooks']
+].map(gclient(LIB_WEBRTC_DIR)).concat([
+  ninja(LIB_WEBRTC_DIR, ['-C', 'trunk/out/Release']),
+  nodegyp_build
+]);
 
-var PROJECT_DIR = process.cwd();
-
-var LIB_DIR = PROJECT_DIR + '/third_party';
-var LIB_WEBRTC_DIR = LIB_DIR + '/libwebrtc';
-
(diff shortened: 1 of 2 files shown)


modeswitch:
There's some overlap with #68. Can you two collaborate on a single patch?

DamonOehlman:
Let's revisit this at a later date.  Packaging up depot tools as a module and dependency for build was as much an experiment as anything else :)

piranna:
> Let's revisit this at a later date. Packaging up depot tools as a module
> and dependency for build was as much an experiment as anything else :)
> 
> I think it makes sense anyway...

Issue #77, opened 2014-02-14
require('wrtc').RTCPeerConnection is undefined


piranna:
As suggested at #76 I went to my code to do `var RTCPeerConnection = require('wrtc').RTCPeerConnection;`, just to my surprise that it was already done in my code, so it was not necesary :-) But, executing my tests on Node.js, I've got the fact that `RTCPeerConnection` is undefined:

```
[piranna@Latitude:~/Dropbox/Proyectos/qunit-cli]
 > bin/qunit-cli ../WebP2P.io/test/2peers.js 
undefined

2 peers
  ✔ Connect two peers to PubNub at the same time

/home/piranna/Dropbox/Proyectos/WebP2P.io/lib/WebP2P.js:78
    var pc = new RTCPeerConnection
             ^
TypeError: undefined is not a function
    at createPeerConnection (/home/piranna/Dropbox/Proyectos/WebP2P.io/lib/WebP2P.js:78:14)
    at HandshakeManager.<anonymous> (/home/piranna/Dropbox/Proyectos/WebP2P.io/lib/WebP2P.js:466:16)
    at HandshakeManager.EventEmitter.emit (events.js:98:17)
    at Connector_PubNub.connector._messageUnpacked (/home/piranna/Dropbox/Proyectos/WebP2P.io/lib/managers/HandshakeManager.js:62:16)
    at Connector_PubNub.<anonymous> (/home/piranna/Dropbox/Proyectos/WebP2P.io/lib/managers/Manager.js:72:21)
    at Connector_PubNub.EventEmitter.emit (events.js:95:17)
    at Array.Connector._message [as 0] (/home/piranna/Dropbox/Proyectos/WebP2P.io/lib/connectors/core/Connector.js:21:10)
    at Array.<anonymous> (/home/piranna/Dropbox/Proyectos/WebP2P.io/node_modules/pubnub/pubnub.js:707:36)
    at each (/home/piranna/Dropbox/Proyectos/WebP2P.io/node_modules/pubnub/pubnub.js:140:15)
    at _connect.xdr.success (/home/piranna/Dropbox/Proyectos/WebP2P.io/node_modules/pubnub/pubnub.js:705:25)
```

As you can see at the top of the trace, that undefined is from executing `console.log(RTCPeerConnection);` statement, showing that, efectively, it's undefined :-(

Executing `console.log(require('wrtc'));` I got

```
{ PeerConnection: [Function: PeerConnection],
[...]

PR #78, opened 2014-02-15
Package.json main pointing to Javascript files instead of bindings

piranna:
This fix issue #77. It also has the fixes done by @wouldgo regarding download of libWebrtc (they should be merged before...).

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "type": "git",
     "url": "http://github.com/js-platform/node-webrtc.git"
   },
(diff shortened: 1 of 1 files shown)


wouldgo:
I didn't notice this thing (in my current development project I'm importing direclty the _index.js_ file).

Good spot! :)

piranna:
Thank you! :-) You should review your project and import just the module, not only to catch this issues :-P but also because it's cleaner to let package.json to route to the module entry point :-)

Issue #80, opened 2014-02-15
Simplify wrappers code structure


piranna:
Currently, the Javascript wrapper class are build using a "private" object that implement all the internal mechanism for the bindings with libWebrtc, and a "public" class that implement the WebRTC Javascript API that only forward requests to the private one. I find this dual-mechanism a non-sense, making the code more complicated and difficult to maintain (I wanted to add EventTarget mechanism but I'm not sure where I should put it) and also consume more memory.

I propose if there's no a good reason to have that architecture to unify both classes and simplify the code, since doing it correctly this doesn't means that private elements could be publicly exposed (in fact, we are already doing it in the private classes and I think it's the reason of wrapping them on the public ones, that I believe it's a really bad thing). If there's no objection, I could be able to do a quick check (for example on RTCPeerConnection object) to see if others agree on this design. I'm not duing it right now since currently I have pending pull requests that I would like them to be accepted (it's better to work with a stable, unified code :-) ).

piranna:
Fixed by issue #84, clossing.

PR #82, opened 2014-02-15
Accessing to module namespace instead of Javascript files

piranna:
As suggested at https://groups.google.com/d/msg/nodejs/RGU0AwTm8cU/Y9y1GppJjA4J

This prevents to have in the future issues like the #77

### DIFF
--- a/examples/bridge.js
+++ b/examples/bridge.js
@@ -1,5 +1,5 @@
 var http = require('http');
-var webrtc = require('../lib/index');
+var webrtc = require('..');
 var ws = require('ws');
 
 var MAX_REQUEST_LENGHT = 1024;
--- a/examples/streamtest/webbridge.js
+++ b/examples/streamtest/webbridge.js
@@ -1,7 +1,7 @@
 var http = require('http');
 var path = require('path');
 
-var webrtc = require('../../lib/index');
+var webrtc = require('../..');
 var express = require('express');
 var socketio = require('socket.io');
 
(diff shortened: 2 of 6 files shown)


piranna:
It has been added the changes for my proposal on issue #80 on this pull-request, I should have move it to a diferent feature branch, sorry. I can do it if required.

Here you have an example of my proposal to unify internal and public classes, in this case done with the PeerConnection one. As a result of it, not only the code is almost 100 lines smaller and gets simpler and cleaner and more easier to maintain and expand (adding EventTarget events will be trivial), but also it will waste less memory and there are no private attributes accesibles from outside, making it more secure and standard compliant. I have updated the methods parameters names to the current ones on the specification. For the native PeerConnection events onicecandidate and ondatachannel I needed to remove the function names due to conflicts that provocated an infinite recursion, but this will be fixed when EventTarget events get implemented.

Opinions? Comments? Suggestions? If you agree, I can keep working on adding the EventTarget events on this and apply the same code scheme to the other public classes.

modeswitch:
Can you split this pull into two?

piranna:
I'll

piranna:
I'll try it.

piranna:
Commits splitted :-) This is what happens when you can't sleep... :-P

wouldgo:
@piranna & @modeswitch: if this PR and the related one are eligible for merging I'll lint the js file after the merging.

piranna:
+1, although I put the extra espaces just for clearance...

PR #86, opened 2014-02-18
List of required packages on Debian-based systems

piranna:


wouldgo:
Yesterday night I read that chromium building machine uses [this](http://src.chromium.org/svn/trunk/src/build/install-build-deps.sh) script as preconfiguration for build phases (and this is kept updated).

Probably our packages requirement is lower but, how about inserting (with _peculiar flags_) as second step in the building phase?

piranna:
It doesn't seems a bad idea, disabling everything the list shouldn't be too big and does it automatically...

jeena:
Perhaps you could add that one needs Java too. I had to do:

```
sudo apt-get install openjdk-6-jdk
```

and then to find the right path to it which in my case was:

```
/usr/lib/jvm/java-6-openjdk-amd64/
```

only there I was able to find the `/include` directory which containes the `jni.h` file. I then exported it in my .bashrc/.zshrc with:

```
export JAVA_HOME=/usr/lib/jvm/java-6-openjdk-amd64/
```

and restarted the terminal.

It took some time because many of the tips on StackExchange let you use either `JAVA_HOME=$(readlink -f /usr/bin/java | sed "s:bin/java::")` which then points to the `jre` directory so you can't see the `/include` dir, or `sudo update-alternatives --config java` which does kind of the same.

piranna:
Why did you need to find the path of jni.h? I don't remember to do it...

jeena:
I don't know but I always got this error:

```
AssertionError: Point $JAVA_HOME or the java_home gyp variable to a directory containing include/jni.h
```

and it wouldn't compile. Perhaps you already had the Java JDK installed and set up from a former project or something?

piranna:
Efectively, I had it installed from a previous install, and at my .bashrc file I had the next line:

```
export JAVA_HOME=/usr/lib/jvm/default-java
```

jeena:
My `/usr/lib/jvm` directory has those under directories:

```
java-1.6.0-openjdk-amd64@
java-6-openjdk-amd64/
java-6-openjdk-common/
java-7-openjdk-amd64/
```

The one with `7` didn't have the `include` dir and neither had the one with `-common` so I used the `-amd64` one.

PR #88, opened 2014-02-20
Toggling tracing

rptr:
I thought that it would be useful to able to hide TRACE calls, so I've added this functionality to node-webrtc.


piranna:
Shouldn't it be better a compile flag and just have traces enabled (or the hability to enable them) on debug mode, and remove it on release?

PR #98, opened 2014-03-02
Support OS X

markandrus:
- Fixes 32/64-bit mismatch
- Adds necessary libraries/frameworks
- `npm test` now works

### DIFF
--- a/bin/build.js
+++ b/bin/build.js
@@ -22,9 +22,10 @@
   process.env.PATH = TOOLS_DEPOT_TOOLS_DIR + ':' + process.env.PATH;
   process.env.GYP_GENERATORS = NINJA;
 
-  if (process.platform === 'darwin')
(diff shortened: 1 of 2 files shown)


pkieltyka:
Btw, I can confirm this is worked on OS X 10.9.1.. I was struggling for npm test to pass until I found this PR. Thank you!

corbanbrook:
I can also confirm that this patch fixes npm install on OS X 10.8.5. all tests pass.

bin/build.js
```diff
@@ -22,6 +22,10 @@
   process.env.PATH = TOOLS_DEPOT_TOOLS_DIR + ':' + process.env.PATH;
   process.env.GYP_GENERATORS = NINJA;
 
+  if (process.platform === 'darwin')
```
modeswitch: Put braces around this branch.

PR #113, opened 2014-04-30
examples/bridge.js now serves static files

mattporritt:
A minor change to help hacks like me with testing.
Altered examples/bridge to also support serving static files to make running the example easier.  More CLI arguments are supported.
Updated documentation and package dependencies to support updated example


modeswitch:
Thanks for the patch!

Issue #43, opened 2013-12-31
Expose DeviceManager and MediaSource/MediaTrack construction, so media can be streamed from devices


kofd:
The cricket API already implements mechanisms for creating streams from attached devices, files, windows, and desktops. This is partially exposed in chrome via getUserMedia(). It would be useful to provide an interface for streaming graphically intensive applications to lightweight peers or providing remote-desktop functionality.

piranna:
Nice suggestion, but I think streams should be generated by other modules.
I believe there's already a Node.js version of getUserMedia(), by the way...

Send from my Samsung Galaxy Note II

kofd:
If you could show me where getUserMedia() is implemented in node, I'd be grateful, I can't find anything other than non-implemented github projects whose goal is to eventually offer that support. 

How would we ensure that the mediasource objects created by the other modules would be compatible with our RTCPeerConnection objects? From what I can tell, unless they also link to our exact build of webrtc and have their own instance of webrtc::PeerConnectionFactoryInterface, it would be impossible to use as a MediaStream object, meaning that we would need to send any video directly through the data channel.

piranna:
We are implementing the WebRTC API on the server side, as far as they
follow the WebRTC or the MediaSource API too there shouldn't be any
problems. Anyway we (and they) could implement support for Node.js stream
API, but this should be a low-priority task or also better an independent
project.

Send from my Samsung Galaxy Note II

kofd:
I guess then I should adjust my implementation of MediaStream and MediaStreamTrack, when PeerConnection unwraps their objects it winds up with our c++ wrapper for the webrtc::MediaStreamInterface object, so it wouldn't be compatible with other libraries. The two libraries would still need to have the c++ libjingle binaries built from the same revision even if it were passing around the same object, but other than that I think I can make it work.

modeswitch:
getUserMedia() is out of scope for this project, I think.

@kofd Is it reasonable to have the MediaStreams API work with node streams? At that point we're dealing with binary data and it's on the application to make sure the stream is usable.

kofd:
[...]

Issue #25, opened 2013-10-26
Support for media stream APIs


modeswitch:
http://www.w3.org/TR/mediacapture-streams/

Create media streams for node that can send and receive video and audio.

piranna:
Seems to me that this is not directly related with WebRTC, but instead on
getUserMedia() and similar, isn't it?

modeswitch:
The PeerConnection API supports adding and removing streams, as well as offering audio and video capability. I would like these to be supported in the node module as well.

I don't see a reason to implement getUserMedia for node.

piranna:
Ok, I agree then.

notedit:
hi  modeswitch:

i just see this 

"
In Chrome 31 and above, a MediaStream from one RTCPeerConnection can be used as the input for another: there's a demo at simpl.info/multi. This can enable more flexible architectures, since it enables a web app to handle call routing by choosing which other peer to connect to
"

in these article http://www.html5rocks.com/en/tutorials/webrtc/infrastructure/

i think it it is easier to do stream  multiplexing.

notedit:
any timeline for this issue?

modeswitch:
Possibly over the next week or so, I think. There are a few use-cases here so I'd like to see this working.

matbeedotcom:
+1
I need to support client stream -> server -> client stream

kofd:
Need the media stream functionality so I've implemented mediastream.h and mediastream.cc, but I'm running into a few issues. I can't get node-webrtc to build right (without my additions) and I'm new to github so I'm not sure how to properly join the project

modeswitch:
@kofd No worries! I've added you as a contributor and assigned this issue to you. I can also try and help you get things compiling; are you on a linux machine?

kofd:
I'm running on Ubuntu 12.04, using ssh on an aws server. I've got it all set up for the most part, but I'm stuck when running:

$ node-gyp --debug rebuild 

The objects binding.o, peerconnection.o, and datachannel.o seem to be created just fine but, but after:

   ...
   CXX(target) Debug/obj.target/webrtc/src/datachannel.o
   SOLINK_MODULE(target) Debug/obj.target/webrtc.node

g++ gives a bunch of "No such file or directory" errors:

   g++: error: ../lib/libwebrtc/trunk/out/Debug/obj/talk/libjingle_peerconnection.a: No such file or directory
   ...
[...]

PR #128, opened 2014-07-09
Add getUserMedia support

markandrus:
This pull request adds initial `getUserMedia` support to node-webrtc. `getUserMedia` has been discussed a little bit (e.g. #25, #43), and I'd like to see it added. For one, it would allow WebRTC clients to move more of their tests out of the browser and into the command-line.

You can use it like

``` js
if (detected === NODE) {
  var wrtc = require('wrtc');
  RTCPeerConnection = wrtc.RTCPeerConnection;
  RTCSessionDescription = wrtc.RTCSessionDescription;
  RTCIceCandidate = wrtc.RTCIceCandidate;
  getUserMedia = wrtc.getUserMedia;
} else if (detected === CHROME) {
  RTCPeerConnection = window.webkitRTCPeerConnection;
  getUserMedia = navigator.webkitGetUserMedia.bind(navigator);
}
```

I tested by making a call between Node.js and Chrome. [See here.](https://gist.github.com/markandrus/c1dad8607de5b7713214)

I'm not sure about functions `runImmediately`, `queueOrRun`, or `_executeNext` in `mediastream.js`, though.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -235,8 +235,9 @@
         'src/set-remote-description-observer.cc',
         'src/peerconnection.cc',
         'src/datachannel.cc',
-#        'src/mediastream.cc',
-#        'src/mediastreamtrack.cc',
+        'src/mediastream.cc',
+        'src/mediastreamtrack.cc',
+        'src/getusermedia.cc'
       ]
     },
     {
--- a/lib/getusermedia.js
+++ b/lib/getusermedia.js
@@ -1,0 +1,16 @@
+var binary = require('node-pre-gyp');
+var path = require('path');
+var binding_path = binary.find(path.resolve(path.join(__dirname,'../package.json')));
+var _webrtc = require(binding_path);
+
+var MediaStream = require('./mediastream');
+
+function getUserMedia(constraints, successCallback, errorCallback) {
+  var stream = _webrtc.getUserMedia(constraints);
+  if (stream === null) {
+    return errorCallback(new Error("Couldn't create MediaStream"));
+  }
+  successCallback(new MediaStream(stream));
+}
+
+module.exports = getUserMedia;
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,3 +1,4 @@
+exports.getUserMedia          = require('./getusermedia');
 exports.RTCIceCandidate       = require('./icecandidate');
 exports.RTCPeerConnection     = require('./peerconnection');
 exports.RTCSessionDescription = require('./sessiondescription');
--- a/lib/mediastream.js
+++ b/lib/mediastream.js
@@ -3,32 +3,6 @@ var EventTarget = require('./eventtarget');
 var MediaStreamTrackEvent = require('./mediastreamtrackevent');
 var MediaStreamTrack      = require('./mediastreamtrack');
 
-
-//MediaStream.prototype._getMS = function _getMS() {
-//  if(!this._ms) {
-//    throw new Error('RTCMediaSteam is gone');
-//  }
-//  return this._ms;
-//};
-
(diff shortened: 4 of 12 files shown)


piranna:
I've review the code and have some questions:
- seems it's audio only and without constraints. What's using as audio input?
- Your demo is only using the stream on browser, how can it be used on Node.js? Could it be possible to pipe the MediaStream or the the MediaTrack objects? How could it be possible to play it and listen on the speakers on Node.js?
- What's missed for adding video support? How could be possible to get a video stream and add it to the PeerConnection object? How could be fetched and piped to another module (for example [fluent-ffmpeg](https://www.npmjs.org/package/fluent-ffmpeg), that support piping of Node.js streams both for input and output)?

modeswitch:
@markandrus Reviewing your PR. Could you rebase to grab the 0.11.13 fixes?

markandrus:
@modeswitch Sorry for the delay. I saw your reply about a week late and have been busy since. I've just updated the PR against the latest on develop.

modeswitch:
@markandrus How much overlap is there with #138?

markandrus:
@modeswitch Excluding the `getUserMedia` changes, they are similar. Some differences I spot, just from glancing over the diffs:
- In `lib/mediastream.js`, #128 gets rid of the `runImmediately` calls within `MediaStream.getId`, `MediaStream.isInactive`, etc. I couldn't tell if these were necessary. Some notes on how the JavaScript wrappers should operate would be helpful.
- In `lib/peerconnection.js`, #128 `PeerConnection.getLocalStreams` and `PeerConnection.getRemoteStreams` delegates to the internal `PeerConnection` methods. #138 returns `[]`.
- In `src/mediastream.cc`, #128 includes a number of mechanical changes to support node 0.11+ as pointed out by @kkoopa, not included in #138.
- In `src/mediastream.cc`, #138 calls `AddRef` and `Release` on the internal `MediaStream`. #128 only sets it to `NULL` in the destructor.
- #128 includes some changes to `examples/streamtest`.

markandrus:
@piranna To answer your questions

> seems it's audio only and without constraints. What's using as audio input?

Right now I think it's choosing a dummy device, and it actually triggers an RTP timeout on my test Asterisk instance. So it's not sending RTP, but it does receive RTP (according to `getStats` from #143, I see packets and bytes received).

> Your demo is only using the stream on browser, how can it be used on Node.js? Could it be possible to pipe the MediaStream or the the MediaTrack objects? How could it be possible to play it and listen on the speakers on Node.js?

I expect this is doable but with more development.

> What's missed for adding video support? How could be possible to get a video stream and add it to the PeerConnection object? How could be fetched and piped to another module (for example fluent-ffmpeg, that support piping of Node.js streams both for input and output)?

Assuming the groundwork to pipe audio gets done, I assume the work would be very similar. WRT fluent-ffmpeg or other projects, I expect that is a ways out and would require much design work.

peteruithoven:
Just checking, would this allow node.js running on a embedded computer (Raspberry Pi like) to stream a webcam video to browsers in a p2p manner? (Using a online webrtc for the signalling etc of course) 
Currently that would require adjusting [Janus](http://janus.conf.meetecho.com/) or the [Flashphoner WebRTC Media & Broadcasting Server](http://flashphoner.com/webrtc-online-broadcasting-from-ip-cameras-and-video-surveillance-systems-using-webrtc-broadcasting-server/), this seems to be much, much easier.

markandrus:
@peteruithoven No, this is not at that level yet.

modeswitch:
At some point, yes. A few patches need to land before that's possible though.

modeswitch:
Native backend won't support media streams at this time.

PR #138, opened 2014-08-24
MediaStreams

blakelapierre:
This gets the streamtest example working. The example shows forwarding MediaStreams for relay/broadcast scenarios.

Mostly I uncommented the MediaStream code that was already in there. I had to make some changes to get it to compile.

I also updated the example to show the original stream that is being sent and the stream that is being rebroadcast.

Enabling TRACING/debug shows that some things may not be cleaned up when a PeerConnection terminates.

Note: There are currently issues with libwebrtc preventing audio tracks from being properly forwarded. It might be possible to hack in a patch to get it working in this case, but according to the webrtc tracker it may be more difficult: https://code.google.com/p/webrtc/issues/detail?id=2192

Sorry for the double pull request. I had rebased my branch to the latest develop and it made the history a little strange. This should be a cleaner request.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -7,3 +7,4 @@ Alex Londeree <alex.londeree@gmail.com>
 Jesús Leganés Combarro "piranna" <piranna@gmail.com>
 Dario Andrei <wouldgo84@gmail.com>
 Matt Porritt <mattp@catalyst-au.net>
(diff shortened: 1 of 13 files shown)


modeswitch:
There's some overlap here with #128.

modeswitch:
Native backend won't support media streams at this time.

PR #169, opened 2015-03-07
Datachannel fix for sending arrayBuffers

vmolsa:


src/datachannel.cc
```diff
@@ -245,14 +245,34 @@ NAN_METHOD(DataChannel::Send) {
...
-    void* data = arraybuffer->GetIndexedPropertiesExternalArrayData();
-    uint32_t data_len = arraybuffer->GetIndexedPropertiesExternalArrayDataLength();
-
-    rtc::Buffer buffer(data, data_len);
-    webrtc::DataBuffer data_buffer(buffer, true);
-    self->_jingleDataChannel->Send(data_buffer);
+  } else if (args[0]->IsArrayBuffer()) {  // Works on node-v0.12.0
+    v8::Local<v8::ArrayBuffer> arraybuffer = v8::Local<v8::ArrayBuffer>::Cast(args[0]);
+    
+    if (arraybuffer->ByteLength()) {
```
modeswitch, line 251: What happens if the caller wants to send an empty buffer?

src/datachannel.cc
```diff
@@ -245,14 +245,34 @@ NAN_METHOD(DataChannel::Send) {
...
     self->_jingleDataChannel->Send(buffer);
-  } else {
-    v8::Local<v8::Object> arraybuffer = v8::Local<v8::Object>::Cast(args[0]);
-    void* data = arraybuffer->GetIndexedPropertiesExternalArrayData();
-    uint32_t data_len = arraybuffer->GetIndexedPropertiesExternalArrayDataLength();
-
-    rtc::Buffer buffer(data, data_len);
-    webrtc::DataBuffer data_buffer(buffer, true);
-    self->_jingleDataChannel->Send(data_buffer);
+  } else if (args[0]->IsArrayBuffer()) {  // Works on node-v0.12.0
```
modeswitch, line 248: This fails on earlier versions of Node. Is there a test you can do that will work on all versions? If not, this might need a pre-processor branch.

PR #170, opened 2015-03-10
Datachannel fix for sending arrayBuffers

vmolsa:


modeswitch:
Thanks for updating this! I'll have a look when I get home tonight.

vmolsa:
Np :) I need those changes for rtc-stream which depends on binary buffer.. 
https://www.npmjs.com/package/rtc-stream

ecoulon:
I have the segmentation fault bug too when sending thousands of buffers. I tried to rebuild with these fixes, the seg fault is gone but i've got this error that happen quickly (about 50-200 buffers sent)
I tried with node >= 0.12.0 and node < 0.12.0

```
node(22846,0x106181000) malloc: *** error for object 0x103070500: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
```

Thanks for your great work!

modeswitch:
@ecoulon It's worth noting that the malloc issue may be covering up the segmentation fault, so it may not be fixed.

@vmolsa Are you able to have a look at the malloc issue?

modeswitch:
@ecoulon Could you link to a gist with your test case?

ecoulon:
I forgot to precise : this bug happen when sending large buffers.
When i send small buffers, i have no error and the seg fault does not occur anymore.

I can reproduce my bug when i set the example/bridge.js to send 200 big buffers.
for example : https://gist.github.com/6d3ee6b6876067b4c939
I change line 115/116

vmolsa:
@ecoulon with node-0.12.0 can you try to comment lines 267 and 268 @ src/datachannel.cc, 

-> // arraybuffer->Neuter();
-> // delete[] static_cast<char*>(content.Data()); 

and rebuild code with "node-gyp build". Is it working? 
But doing so it will start to collect memory and not freeing it?

ecoulon:
Hmm with these two lines commented i have 3 different results (seems to be random) : 
- the most often i have the same error
- i had the seg fault a few times
- it works a few times too with no error

vmolsa:
Thanks, I created gist for sending multiple buffers with different times / sizes. And yes it's failing almost every time.. 

https://gist.github.com/vmolsa/84aab2974132d189574e

vmolsa:
Datachannel max buffer size on node is 66528 bytes and on browser it's 73905 bytes (why?) without splitting input buffers and cause of that splitting some bytes will be lost..

Most of segfaults are coming from webrtc thread. rtc::Buffer() allocates new memory segment each time and copies data from string / arraybuffer to that segment and webrtc thread releases that buffer after it's not needed. So the problem is not in node-webrtc module?.. 

@modeswitch can you update libwebrtc?

node-v0.12.0
node(57067,0x1016fc000) malloc: **\* error for object 0x1: pointer being freed was not allocated
**\* set a breakpoint in malloc_error_break to debug
0: SIGABRT [0x7fff9811f41d] thread_start
1: SIGABRT [0x7fff981211e5] _pthread_body
2: SIGABRT [0x7fff98121268] _pthread_body
3: SIGABRT [0x10340b9c6] rtc::Thread::PreRun(void_)
4: SIGABRT [0x10340bd3d] rtc::Thread::ProcessMessages(int)
5: SIGABRT [0x10335a000] cricket::DataChannel::OnMessage(rtc::Message_)
6: SIGABRT [0x10330c10f] webrtc::DataChannel::SendQueuedDataMessages()
7: SIGABRT [0x7fff9a6ad937] nano_size
8: SIGABRT [0x7fff8f102b53] abort
9: SIGABRT [0x10131bd01] function()

node-v0.11.16
0: SIGSEGV [0x106fb03060a2] function()
1: SIGSEGV [0x10042b23f] v8::internal::Runtime_TypedArrayInitializeFromArrayLike(int, v8::internal::Object*_, v8::internal::Isolate_)
2: SIGSEGV [0x1002334ad] v8::internal::Factory::NewJSArrayBuffer()
3: SIGSEGV [0x10026853d] v8::internal::Heap::CollectGarbage(v8::internal::GarbageCollector, char const_, char const_, v8::GCCallbackFlags)
4: SIGSEGV [0x100268c3f] v8::internal::Heap::PerformGarbageCollection(v8::internal::GarbageCollector, v8::GCCallbackFlags)
5: SIGSEGV [0x10024a787] v8::internal::GlobalHandles::PostGarbageCollectionProcessing(v8::internal::GarbageCollector)
6: SIGSEGV [0x10024ba34] v8::internal::GlobalHandles::Node::PostGarbageCollectionProcessing(v8::internal::Isolate*)
7: SIGSEGV [0x103a89707] node::ObjectWrap::WeakCallback(v8::WeakCallbackData<v8::Object, node::ObjectWrap> const&)
8: SIGSEGV [0x103a8a688] node_webrtc::DataChannel::~DataChannel()
9: SIGSEGV [0x103a8a648] node_webrtc::DataChannel::~DataChannel()
10: SIGSEGV [0x103a993ff] rtc::RefCountedObjectwebrtc::DataChannelProxy::Release()
11: SIGSEGV [0x103a9942e] rtc::RefCountedObjectwebrtc::DataChannelProxy::~RefCountedObject()
12: SIGSEGV [0x103a9a5bb] webrtc::DataChannelProxy::~DataChannelProxy()

lib/datachannel.js
```diff
@@ -41,6 +41,11 @@ function RTCDataChannel(internalDC) {
         return this.RTCDataStates[state];
       }
     },
+    'bufferedAmount': {
+      get: function getReadyState() {
```
modeswitch, line 45: getReadyState is the wrong name for this function

src/datachannel.h
```diff
@@ -60,9 +60,9 @@ class DataChannel
   };
 
   enum AsyncEventType {
-    MESSAGE = 0x1 << 0, // 1
-    ERROR = 0x1 << 1, // 2
-    STATE = 0x1 << 2, // 4
+    MESSAGE = 1,
```
modeswitch, line 63: Why were these changed?

src/peerconnection.cc
```diff
@@ -649,6 +849,105 @@ NAN_SETTER(PeerConnection::ReadOnly) {
   INFO("PeerConnection::ReadOnly");
 }
 
+#define USE_BACKTRACE 1
```
modeswitch, line 852: Why is this enabled by default?

src/peerconnection.h
```diff
@@ -94,38 +94,30 @@ class PeerConnection
...
-
-    ERROR_EVENT = CREATE_OFFER_ERROR | CREATE_ANSWER_ERROR |
-                  SET_LOCAL_DESCRIPTION_ERROR | SET_REMOTE_DESCRIPTION_ERROR |
-                  ADD_ICE_CANDIDATE_ERROR,
-    SDP_EVENT = CREATE_OFFER_SUCCESS | CREATE_ANSWER_SUCCESS,
-    VOID_EVENT = SET_LOCAL_DESCRIPTION_SUCCESS | SET_REMOTE_DESCRIPTION_SUCCESS |
-                 ADD_ICE_CANDIDATE_SUCCESS | GET_STATS_SUCCESS,
-    STATE_EVENT = SIGNALING_STATE_CHANGE | ICE_CONNECTION_STATE_CHANGE |
-                  ICE_GATHERING_STATE_CHANGE
+    CREATE_OFFER_SUCCESS = 1,
```
modeswitch, line 97: I think this should be fine, as long as you unfolded the combined events from the old code.

src/peerconnection.cc
modeswitch, line 852: Also, please explain this a little.

src/peerconnection.cc
```diff
@@ -95,130 +180,233 @@ void PeerConnection::Run(uv_async_t* handle, int status)
...
+            v8::Local<v8::Value> argv[1];
+
+            argv[0] = NanNew<Uint32>(data->state);
+
+            NanMakeCallback(pc, callback, 1, argv);
+          }
+        }
+        
+        break;
+      case PeerConnection::NOTIFY_ADD_STREAM:
```
modeswitch, line 362: This can be removed entirely, along with NOTIFY_REMOVE_STREAM.

src/peerconnection.cc
```diff
@@ -34,27 +34,112 @@ Persistent<Function> PeerConnection::constructor;
...
+          v8::Local<v8::Object> option = v8::Local<v8::Object>::Cast(option_value);
+          v8::Local<v8::Value> DtlsSrtpKeyAgreement = option->Get(NanNew("DtlsSrtpKeyAgreement"));
+          
+          if (!DtlsSrtpKeyAgreement.IsEmpty() && DtlsSrtpKeyAgreement->IsFalse()) {
+            enableSecure = false;
+          }
+        }
+      }
+    }
+/*
```
modeswitch, line 107: This should be removed instead of commented out. If we need it again, it's in the repo history.

src/datachannel.cc
```diff
@@ -239,21 +259,60 @@ NAN_METHOD(DataChannel::Send) {
...
+      v8::Local<v8::ArrayBuffer> arraybuffer;
+      
+      if (args[0]->IsArrayBuffer()) {
+        arraybuffer = v8::Local<v8::ArrayBuffer>::Cast(args[0]);
+      } else {
+        v8::Local<v8::ArrayBufferView> view = v8::Local<v8::ArrayBufferView>::Cast(args[0]);
+        arraybuffer = view->Buffer();
+      }
+      
+      if (!arraybuffer.IsEmpty()) {
```
modeswitch, line 284: We should allow sending an empty buffer, no?

modeswitch:
Can you elaborate on the problem this patch is meant to fix? It's a little difficult given the size of the patch to determine whether it addresses the original issue.

.travis.yml
```diff
@@ -7,6 +7,7 @@ language: node_js
 node_js:
   - "0.10"
   - "0.11"
+  - "0.12"
```
modeswitch, line 10: This was pulled in from another patch, so you can revert it here.

lib/datachannel.js
modeswitch, line 45: This was pulled in from another patch, so you can revert the implementation for bufferedAmount. If you update your local copy it will be there.

src/datachannel.cc
vmolsa, line 284: It will send empty buffers.

src/peerconnection.cc
vmolsa, line 107: It will gain better performance by using switch cases.

src/peerconnection.cc
vmolsa, line 852: http://www.gnu.org/software/libc/manual/html_node/Backtraces.html

vmolsa:
I'm currently working with another webrtc fork. These changes will fix some of problems. 
Meaning that this project should be more compatible with webrtc API. 

https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection

src/datachannel.cc
modeswitch, line 284: Right, this is a test on the handle. Ignore.

modeswitch:
@vmolsa Why was this closed?

Issue #168, opened 2015-03-01
Prebuilt binaries


guymguym:
Hi

Great project!

I think it would be even more usable if it would provide prebuilt binaries.

Build from source takes a long time and as such is less attractive to embed in existing projects.

I think this also helped node-webkit (https://github.com/nwjs/nw.js) gain popularity - if I had to build it from source in every project it would be a real pain.

What do you think?

modeswitch:
Yeah, I will investigate compiled binaries for some common platforms.

piedshag:
@modeswitch I have placed a 15 dollar bounty on this issue. I hope it is of assistance.

modeswitch:
@fbodz I don't understand what you mean. Is this a Github feature?

rosskukulinski:
@guymguym AND @fbodz -- do you have any preferences on initial binaries to support?  I was thinking OSX & Ubuntu x86_64

piedshag:
@modeswitch There is a service called bountysource which allows you to place bounties on Github issues. Here is the url https://www.bountysource.com/issues/9082012-prebuilt-binaries. @rosskukulinski  I think OSX and Ubuntu x86_64 would be a great start. Thanks. Let the race begin!

rosskukulinski:
Working on a fork with automated pre-compiled binaries.

piedshag:
@rosskukulinski Have you logged in as a developer on bounty source. Therefore i can release the bounty when the fork is merged.

guymguym:
I am eventually interested in Windows, but this is still an open issue #42.
OSX is a good start as well.
:+1:

rosskukulinski:
@fbodz I'll gladly accept the bounty and then donate to charity.  I'm more interested in speeding up our own build->test->deploy process.

piedshag:
@rosskukulinski Sounds great. When the prebuilt binaries are released and the issue is closed. I will be more than happy to release the bounty. Do you have any idea when you think the binaries will be completed.

rosskukulinski:
@modeswitch - the test for this will be to verify that the next time you tag a new version and push -- does the automated build work correctly.

modeswitch:
@rosskukulinski I'll publish when I have a sec.

FWIW, I was using node-pre-gyp before but it was more trouble that it was worth.

rosskukulinski:
Well hopefully I haven't made things worse :-/.  "Works on my machine" :)

modeswitch:
[...]

PR #173, opened 2015-03-28
Add prebuilt binaries for Linux and OSX

rosskukulinski:
Travis automatically builds binaries for linux x86_64.  OSX Builds can/will be published manually. Refs #168.

You can test the linux x86_64 and OSX pre-built binaries by using 
`"wrtc": "git+ssh://git@github.com:yodlr/node-webrtc.git#prebuilt"` in your package.json

Last successful build from our branch was here: https://travis-ci.org/yodlr/node-webrtc/jobs/56176259

The current published binaries include:

```
[~/dev/yodlr/node-webrtc (prebuilt_bin)]$ s3-cli ls --recursive s3://node-webrtc/
Fri Mar 27 2015 18:31:23 GMT-0700 (PDT) 1051400 wrtc/v0.0.49/Release/node-v0.11.16-linux-x64.tar.gz
Fri Mar 27 2015 18:28:31 GMT-0700 (PDT) 1562226 wrtc/v0.0.49/Release/node-v11-darwin-x64.tar.gz
Fri Mar 27 2015 18:30:58 GMT-0700 (PDT) 1050977 wrtc/v0.0.49/Release/node-v11-linux-x64.tar.gz
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,52 @@ before_install:
   - sudo apt-get install -y python libnss3-dev libncurses5-dev libssl-dev libexpat-dev
   - npm install -g grunt-cli
   - npm install node-gyp -g
+  # get commit message
+  - COMMIT_MESSAGE=$(git show -s --format=%B $TRAVIS_COMMIT | tr -d '\n')
+  # put local node-pre-gyp on PATH
+  - export PATH=./node_modules/.bin/:$PATH
+  # install aws-sdk so it is available for publishing
+  - npm install aws-sdk
+  # figure out if we should publish
+  - PUBLISH_BINARY=false
+  # if we are building a tag then publish
+  - if [[ $TRAVIS_BRANCH == `git describe --tags --always HEAD` ]]; then PUBLISH_BINARY=true; fi;
+  # or if we put [publish binary] in the commit message
+  - if test "${COMMIT_MESSAGE#*'[publish binary]'}" != "$COMMIT_MESSAGE"; then PUBLISH_BINARY=true; fi;
+  - platform=$(uname -s | sed "y/ABCDEFGHIJKLMNOPQRSTUVWXYZ/abcdefghijklmnopqrstuvwxyz/")
+  # If version is 0.9 do not publish to avoid duplicated binary error, caused by 0.9 and 0.10
+  # returning same v number in node.js.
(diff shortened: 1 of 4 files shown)


piedshag:
@rosskukulinski Fabulous. :+1:

guymguym:
@rosskukulinski Great! 

suggestions:
1. I think npm install should generally pull the prebuilt binary, and only do build-from-source if manually running bin/build.js. I hope this will be convenient for developers, but this is surely better for users of the project?
2. it makes bit more sense to me to name the files node-webrtc-... instead of just the node-... prefix because it's not identifiable this way.

rosskukulinski:
Hi @guymguym -

1) It _should_ pull down the prebuilt binary first, then build from source if no-matching binary is found.  Notice that the npm install command was modified to run `"install": "node-pre-gyp install --fallback-to-build",` [here](https://github.com/js-platform/node-webrtc/pull/173/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R50).  If this is NOT the case, please let me know!  Either A) I made a mistake or B) I haven't created a build for your platform (yet).

2) Ah, I see what you mean.  Well, the binaries get pulled down automatically by npm and put within `node_modules/wrtc/build/wrtc/v{version}/{configuration}/{node_abi}-{platform}-{arch}/` which based on my research seems to be the 'standard' way with Nodejs native modules.  Since the binary is buried under the wrtc path, I think this is ok.
See other examples:
- [node-zipfile](https://github.com/mapbox/node-zipfile/blob/master/package.json#L36)
- [node-sqlite3](https://github.com/mapbox/node-sqlite3/blob/master/package.json#L12)
- https://github.com/mapbox/node-pre-gyp/wiki/Modules-using-node-pre-gyp

Do you have a use for the binaries in some other manner?  I'm cool with changing this if there's a reason.

guymguym:
@rosskukulinski thanks for the explanations. 
1. I will give it a try
2. I just downloaded manually from s3 location you wrote so it was looking odd, but if done automatically as part of the build then it should be fine.

rosskukulinski:
@guymguym it should all be handled by npm magically behind the scenes.  0.0.49 has Linux & OSX, 0.0.50 only has OSX.

I'm just waiting for @modeswitch to try another version bump (https://github.com/js-platform/node-webrtc/issues/168#issuecomment-87830148) to see if we get automated Linux for 0.0.51.

Issue #171, opened 2015-03-23
malloc: *** error for object 0x100000000: pointer being freed was not allocated


guymguym:
hi

I see this issue like discussed in #170 

```
node(46307,0x104516000) malloc: *** error for object 0x100000000: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

though I use a different test - 
https://gist.github.com/guymguym/728bf41ab38466ccca4f (see usage in gist comment).  
when the two peers are chrome/firefox it works well (though with 100% cpu for just ~5 MB/s on Macbook Air :snail: ). 
but any wrtc client fails pretty quickly.

this occurs on OSX 10.10.2 with latest wrtc version which is 0.0.49 in npm (which is not the same as the github release numbers ?!).

guymguym:
@modeswitch Hi, 
This issue is quite the killer, as it fails every time.
Do you see it reproducing too? 
Does pull-request #170 offers any fix around these memory allocation issues?
Any idea?
Thanks!

modeswitch:
@guymguym I'm having a look now.

guymguym:
:+1:

modeswitch:
@guymguym I'm seeing a crash. I'm also seeing that only some of the buffers are arriving in the browser.

guymguym:
by "some buffers are arriving" do you mean that when you run browser vs. browser you see "BAD SEQUENCE" prints? because the code tries to check that all packets arrive in order...
or you mean browser vs. node?

modeswitch:
No, I mean only some of the buffers are printed to the console. I'm using the modified bridge from #170.

guymguym:
I didn't follow which prints you mean, so not sure if I can assist.

BTW you maintain a copy of webrtc repo in [libwebrtc](https://github.com/js-platform/libwebrtc). Doesn't it make it hard to sync with changes and avoid collisions? Wasn't it possible to use the original repo and just make a custom build and bridge for node?

modeswitch:
The buffers are printed to the console as they are received.

I was originally using upstream libjingle, but it has many dependencies and is therefore difficult to build for many people. The fork I have is missing the implementation for media streams, which dramatically reduces the number of external dependencies.

guymguym:
Hey @modeswitch just wondering if you got any lead on that?

[...]

PR #186, opened 2015-04-04
bwtest improved and added to npm test

guymguym:
- when running directly as main script will use minimist args, so can
  run like: node test/bwtest —packetCount 1000
- npm test runs 3 short bwtest variants (with tape). the current
  variants are: default, no buffering, unordered and unreliable.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -7,7 +7,7 @@
   "eqeqeq": true,
   "immed": true,
   "indent": 2,
-  "latedef": true,
+  "latedef": false,
   "newcap": true,
   "noarg": true,
   "noempty": true,
--- a/.jshintrc
+++ b/.jshintrc
@@ -7,7 +7,7 @@
   "eqeqeq": true,
   "immed": true,
   "indent": 2,
-  "latedef": false,
+  "latedef": "nofunc",
   "newcap": true,
   "noarg": true,
   "noempty": true,
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "p2p",
     "peer"
   ],
-  "version": "0.0.52",
+  "version": "0.0.53",
   "author": "Alan K <ack@modeswitch.org> (http://blog.modeswitch.org)",
   "homepage": "http://js-platform.github.io/node-webrtc/",
   "bugs": "https://github.com/js-platform/node-webrtc/issues",
--- a/test/all.js
+++ b/test/all.js
@@ -1,3 +1,4 @@
 require('./create-offer');
 require('./sessiondesc');
 require('./connect');
+require('./bwtest').tape();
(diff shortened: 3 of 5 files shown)


guymguym:
the build fails during bwtest (as expected) because we still don't have the webrtc fix from upstream as discussed in #171.

```
*** glibc detected *** node: free(): invalid pointer: 0x00007f5dbc054c90 ***
```

guymguym:
@modeswitch I'm trying to update this pull-request to trigger a rebuild but for some reason it doesn't work. I added two more commits (one to bump the version) to the pull-request branch (https://github.com/guymguym/node-webrtc/commits/develop) but it doesn't seem to impress github and this pull-req remains the same... Any ideas?

modeswitch:
@guymguym Looks like it's rebuilding now. I can also trigger it manually if we need.

guymguym:
thanks, I found how to manually re-open the pull-request from my fork, and it did the trick

guymguym:
ok it works for node 0.10 but running a bandwidth test fails for 0.11 and 0.12, which I haven't tested on my env. might be related to https://github.com/js-platform/node-webrtc/pull/185.

For node 0.11 the error is:

```
# bwtest default
SENDING: sent 0 received 0. congestion #0 0.000 seconds. took 0.000 seconds. bandwidth NaN KB/s. 
Segmentation fault
npm ERR! Test failed.  See above for more details.
```

For node 0.12 the error is:

```
# bwtest default
SENDING: sent 0 received 0. congestion #0 0.000 seconds. took 0.001 seconds. bandwidth 0 KB/s. 
FATAL ERROR: v8::ArrayBuffer::Externalize ArrayBuffer already externalized
Aborted
npm ERR! Test failed.  See above for more details.
```

modeswitch:
Having a look now.

guymguym:
In bwtest I used the same ArrayBuffer to send all the packets (which is valid), but it fails because Externalize() is only allowed to be called once, and once it is externalized, it becomes the responsibility of the caller to grab the pointer and release it.

See the full discussion here: https://groups.google.com/forum/#!topic/v8-users/whHPAIhfMu8

They give an example of code that manages the pointer along with the ArrayBuffer using Get/SetAlignedPointerFromInternalField. 
https://code.google.com/p/chromium/codesearch#chromium/src/gin/array_buffer.cc&l=93

modeswitch:
Cool. I'm just fixing another issue with the way the binding is loaded. Can you make a pull request to fix this?

guymguym:
Sure

modeswitch:
@guymguym Btw, I'm having issues with the debug build (assertion failure in physicalsocketserver.cc). Did you and rossk find a workaround for that?

guymguym:
we didn't see that. we only saw debug prints caused by kSctpMtu=1200: https://github.com/js-platform/node-webrtc/issues/171#issuecomment-88639805

PR #188, opened 2015-04-06
Fix arguments to createOffer

### LINKED ISSUE #178: Wrong usage of "arguments"
In `lib/peerconnection.js`:

```
javascript
  this.createOffer = function createOffer(successCallback, failureCallback, options) {
    options = options || {};

    // FIXME: complain if onError is undefined
    queueOrRun({
      func: 'createOffer',
      args: [options],
      wait: true,
      onSuccess: function(sdp) {
        successCallback.call(this, new RTCSessionDescription({type: 'offer', sdp: sdp}));
      },
      onError: failureCallback
    });

    return pc.createOffer(arguments);
  };
```

That is wrong. `arguments` is a special object. It should be called as follows:

```
return pc.createOffer.apply(pc, arguments);
```

or even easier and better:

```
return pc.createOffer(successCallback, failureCallback, options);
```

rosskukulinski:
Fixes #178

### DIFF
--- a/lib/peerconnection.js
+++ b/lib/peerconnection.js
@@ -216,7 +216,7 @@ function RTCPeerConnection(configuration, constraints) {
       onError: failureCallback
     });
 
-    return pc.createOffer(arguments);
+    return pc.createOffer(successCallback, failureCallback, options);
   };
 
   this.createAnswer = function createAnswer(successCallback, failureCallback, options) {


rosskukulinski:
Build (will) error out due to BW test.  See #189 where I've disabled that test until its fixed.

modeswitch:
It should still publish builds for v0.10, no? I'd rather not disable the failing test.

rosskukulinski:
uh, yes, it should @modeswitch, but 0.11, 0.12, and iojs will all fail: https://travis-ci.org/yodlr/node-webrtc/builds/57358271

modeswitch:
It should fail until @guymguym has his fix merged, no? That's why the tests are run automatically by Travis.

guymguym:
hey, the test is innocent, it's just uncovering bugs :)
I actually tried to run bwtest with allocating new buffer for each send and still see that 0.11 and 0.12 fail. Handling ArrayBuffers is still not solid for 0.11 and 0.12 around https://github.com/js-platform/node-webrtc/blob/develop/src/datachannel.cc#L249-L269 . 

I will probably won't get to it in the next week or so, so I will change bwtest to allocate new buffer for now to avoid repeated Externalize() issues, and maybe @janza can give an advice as a followup to #185 since bwtest wasn't there to check when he authored.

Clearly if it's more important for the build to pass, you can just disable bwtest.

guymguym:
@rosskukulinski try merging after #191, though on my local tests it still fails, but somehow the build on linux passed..

Issue #190, opened 2015-04-06
bwtest fails on node 0.12 and 0.11


guymguym:
for some reason this only occurs when running test/all.js and not a single test/bwtest.js and always on the second run of (`# bwtest no buffering`) while destructing the data channel. in this example with node 0.12:

```
$ lldb -- node test/all.js 
...
# bwtest default
SENDING: sent 0 received 0. congestion #0 0.000 seconds. took 0.001 seconds. bandwidth 0 KB/s. 
SENDING: sent 100 received 28. congestion #0 0.000 seconds. bufferedAmount 950272. took 0.039 seconds. bandwidth 11487 KB/s. 
SENDING: sent 200 received 131. congestion #12 0.081 seconds. bufferedAmount 999424. took 0.168 seconds. bandwidth 12476 KB/s. 
SENDING: sent 300 received 257. congestion #31 0.181 seconds. bufferedAmount 540672. took 0.297 seconds. bandwidth 13845 KB/s. 
SENDING: sent 400 received 370. congestion #50 0.279 seconds. bufferedAmount 327680. took 0.421 seconds. bandwidth 14062 KB/s. 
SEND DONE! sent 500 received 436. congestion #60 0.332 seconds. bufferedAmount 917504. took 0.493 seconds. bandwidth 14150 KB/s. 
RECEIVE DONE! sent 500 received 500. congestion #60 0.332 seconds. took 0.573 seconds. bandwidth 13962 KB/s. 
ok 51 bwtest check for error
# bwtest no buffering
SENDING: sent 0 received 0. congestion #0 0.000 seconds. took 0.000 seconds. bandwidth NaN KB/s. 
SENDING: sent 100 received 69. congestion #12 0.063 seconds. bufferedAmount 294912. took 0.092 seconds. bandwidth 12000 KB/s. 
SENDING: sent 200 received 138. congestion #24 0.121 seconds. bufferedAmount 884736. took 0.170 seconds. bandwidth 12988 KB/s. 
SENDING: sent 300 received 263. congestion #51 0.269 seconds. bufferedAmount 425984. took 0.339 seconds. bandwidth 12413 KB/s. 
SENDING: sent 400 received 333. congestion #63 0.328 seconds. bufferedAmount 1015808. took 0.418 seconds. bandwidth 12746 KB/s. 
Process 49012 stopped
[...]

PR #191, opened 2015-04-06
bwtest workaround over Externalize, node 0.11 ArrayBuffer path like 0.12

guymguym:


guymguym:
#190 will still fail though so the build fail is expected

PR #196, opened 2015-04-11
fix #194 install simple-peer on dev mode with --no-optional

### LINKED ISSUE #194: Simple-Peer devDependency tries to compile wrtc as optionalDependency
I added the devDependency on simple-peer in bwtest as it really simplifies flow for small tests and such.

But now @feross added simple-peer an [optionalDependency](https://docs.npmjs.com/files/package.json#optionaldependencies) back on node-webrtc :) https://github.com/feross/simple-peer/commit/319a862d425965838b2501e2421e97f4aca8f5b2

So now when building wrtc, the devDependency on simple-peer is causing it to install, and then it attempts to install it's optional dependency on wrtc which is ridiculous when we actually in developer mode of wrtc.

The only solution I think of if in simple-peer the dependency would be devDependency, but there is not npm support for devOptionalDependency, so not sure how to fix. Ideas?

guymguym:


guymguym:
maybe wait with this pull-request to see how this https://github.com/feross/simple-peer/issues/14 turns out.

modeswitch:
@guymguym Let me know when this is OK to merge.

guymguym:
@modeswitch I updated the pull to use simple-peer 5.0 which now supports wrtc option, so you can review the change and merge.

modeswitch:
Look fine to me. The test will still fail due to a known issue with data channels.

PR #215, opened 2015-07-07
Link statically built boringssl over nss

jonathanfrawley:
- Nss on ubuntu 14.04 has no compatible ciphers with firefox 39
- BoringSSL has compatible ciphers so this fixes the bridge example for firefox
- Requires commit f632d8d3203af25bb8bdd7900a8d7875085c9143 of libwebrtc
- See pull request on libwebrtc for more details : https://github.com/js-platform/libwebrtc/pull/12

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "minimist": "0.0.8"
   },
   "scripts": {
-    "install": "node-pre-gyp install --fallback-to-build",
+    "install": "node-pre-gyp install --build-from-source",
     "test": "node test/all.js"
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -49,7 +49,7 @@
     "minimist": "0.0.8"
   },
   "scripts": {
-    "install": "node-pre-gyp install --fallback-to-build",
+    "install": "node-pre-gyp install --build-from-source",
```
modeswitch: I don't think this should be changed.

jonathanfrawley: Reverted that, forgot to remove before committing.

PR #247, opened 2015-12-18
Add g++ as a pre-requisite (apt dependency) on README.md

nihey:
I'm not quite familiar with `node-gyp` and native addons to Node.js.

But @hackedbellini was trying to install `npm i wrtc` on a brand new
Ubuntu Trusty (14.04) and was having some issues on installation,
it seemed as `g++` was missing, so the whole installation failed.

Another evidence that `g++` is needed here is that it is listed on
the [`Dockerfile`](https://github.com/js-platform/node-webrtc/blob/develop/Dockerfile) dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Some open issues have bounties assocaited with them. Once you patch is merged, y
 
 * python2
 * git
+* g++
 * pkg-config
 * libncurses-devel
 * libssl-devel
@@ -38,7 +39,7 @@ Some open issues have bounties assocaited with them. Once you patch is merged, y
 
 To install prerequisites with the apt-get package manager,
 
-`apt-get install python2.7 git-all pkg-config libncurses5-dev libssl-dev libnss3-dev libexpat-dev  `
+`apt-get install python2.7 git-all g++ pkg-config libncurses5-dev libssl-dev libnss3-dev libexpat-dev`
 
 ### CentOS/Fedora/RHEL
 


nihey:
This is kind of conflicting with https://github.com/js-platform/node-webrtc/pull/238 , but as it was never merged nor updated, I thought it was a good idea to put it here.

@hackedbellini may be able to give the exact error log caused by the missing `g++` on monday.

bellini666:
As @nihey said, I had problems installing `wrtc` on a fresh Ubuntu 14.04 installation. I installed everything that the `README.md` said I should, but I was only able to complete the installation after I installed `g++`. The error itself was also really hard to spot, since it produced a lot of side-effects instead of failing immediately.

Just a small correction, the package that required it was `node-pre-gyp`, not `node-gyp`.

If you need I can provide some logs to give the exact error, but it should also be really easy to reproduce the issue by simply temporarily removing `g++` from `PATH`.

modeswitch:
`g++` isn't a direct dependency, so it shouldn't be listed here. `node-pre-gyp` is listed as a dependency in the `package.json`.

bellini666:
@modeswitch yes I agree, but even though it is not a direct dependency, it is something you will need to be able to successfully install `wrtc`. If every time you want to install something you need on `node` you have to manually analyze each of its dependencies (and o course, the dependencies dependencies and so on), you will go crazy. Lots of packages in the dependency tree.

PR #263, opened 2016-03-17
Remove calls to Externalize and Neuter

### LINKED ISSUE #262: Arraybuffers are "Externalized" when sending, but are not "un-externalized"
I can send an arraybuffer to a peer just fine - but If I attempt to send the same array buffer to another peer immediately afterward, I get the following error:

```
FATAL ERROR: v8::ArrayBuffer::Externalize ArrayBuffer already externalized
Aborted (core dumped)
```

To "fix" this I simply have to re-create a new ArrayBuffer and send that to each additional client. But it'd be nice if you could reuse the same ArrayBuffer. Does it need to be "un-externalized" before the send function completes?  Perhaps it needs to be 'rewound'?

modeswitch:
Fix #262.

### DIFF
--- a/src/datachannel.cc
+++ b/src/datachannel.cc
@@ -258,7 +258,6 @@ NAN_METHOD(DataChannel::Send) {
       arraybuffer = view->Buffer();
     }
 
-    v8::ArrayBuffer::Contents content = arraybuffer->Externalize();
     rtc::Buffer buffer(content.Data(), content.ByteLength());
 
 #else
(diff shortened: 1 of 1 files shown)


modeswitch:
Review/merge when you get a sec.

src/datachannel.cc
```diff
@@ -258,7 +258,6 @@ NAN_METHOD(DataChannel::Send) {
       arraybuffer = view->Buffer();
     }
 
-    v8::ArrayBuffer::Contents content = arraybuffer->Externalize();
     rtc::Buffer buffer(content.Data(), content.ByteLength());
```
markandrus, line 261: This line fails because we no longer define `content`. I am not sure how to get a hold of the `v8::ArrayBuffer::Contents` or the underlying data without `Externalize`—I need to read up on ArrayBuffer

modeswitch:
Wrong fix :)

PR #264, opened 2016-03-18
added test for sending an arraybuffer multiple times

bfishman:
as expected, this currently fails at the new test with the error `FATAL ERROR: v8::ArrayBuffer::Externalize ArrayBuffer already externalized`


Globik:
And what now? With node.js 6 datachannel.cc cannot compile.

markandrus:
See #338

PR #271, opened 2016-05-09
Update nan to fix deprecation warnings

caseywebdev:
Node 6 dumps a heap of deprecation warnings when this lib is included:

```
(node) v8::ObjectTemplate::Set() with non-primitive values is deprecated
(node) and will stop working in the next major release.
==== JS stack trace =========================================
Security context: 0x16917dac9fa9 <JS Object>#0#
    1: .node [module.js:568] [pc=0x43fc98a3704] (this=0x206b7dfdf3e9 <an Object with map 0x31b8ee17609>#1#,module=0x34b4d60d4d59 <a Module with map 0x31b8ee17be1>#2#,filename=0x34b4d60d4d21 <String[61]: /code/build/wrtc/v0.0.59/Release/node-v48-linux-x64/wrtc.node>)
...
```

It also warns that it will not work altogether in node 7.

[Another project with the same issue](https://github.com/Automattic/node-canvas/issues/758) narrowed it down to an outdated version of nan. It'd be great to be able to merge this and cut a release to fix those noisy warnings that are blowing up my app logs.

Edit: It looks like the old version of `ws` that is used for testing also spews the deprecation warning as it uses `nan@1.0.0`, so I updated that as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "fs-extra": "^0.18.0",
-    "nan": "^2.1.0",
+    "nan": "^2.3.2",
     "node-gyp": "^3.0.3",
     "node-pre-gyp": "0.6.x",
     "node-static-alias": "^0.1.2",
@@ -45,7 +45,7 @@
     "grunt-tape": "~0.0.2",
     "simple-peer": "^5.0.0",
     "tape": "~2.4.2",
-    "ws": "~0.4.31",
+    "ws": "^1.1.0",
     "node-static": "^0.7.3",
     "minimist": "0.0.8"
   },
--- a/src/datachannel.cc
+++ b/src/datachannel.cc
@@ -351,13 +351,10 @@ void DataChannel::Init(Handle<Object> exports) {
   Local<FunctionTemplate> tpl = Nan::New<FunctionTemplate>(New);
   tpl->SetClassName(Nan::New("DataChannel").ToLocalChecked());
   tpl->InstanceTemplate()->SetInternalFieldCount(1);
-  tpl->PrototypeTemplate()->Set(Nan::New("close").ToLocalChecked(),
-    Nan::New<FunctionTemplate>(Close)->GetFunction());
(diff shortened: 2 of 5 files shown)


caseywebdev:
Whelp it appears to still throw those deprecation warnings in the node 6 travis build. They must be coming from somewhere in addition to `nan`.

markandrus:
Thanks for the report @caseywebdev—perhaps there is another place in the code that's calling `v8::ObjectTemplate::Set`?

caseywebdev:
Yea I found it. I have extremely limited experience with v8 but I've narrowed it down to the 15 instances of

``` c++
tpl->PrototypeTemplate()->Set(...)
```

Here's the commit in `Nan` that fixed the issue for them: https://github.com/nodejs/nan/commit/a90951e9ea70fa1b3836af4b925322919159100e

I'm trying different variations of this til it works ;) If you figure it out please let me know.

caseywebdev:
Got it! The tests are passing and the warning is gone now!

markandrus:
Great work @caseywebdev!

caseywebdev:
🎆

caseywebdev:
Would you mind cutting a release soon-ish? These deprecation warnings are extremely noisy in the logs and I'd love to be rid of them.

markandrus:
Yes, let me check in with @modeswitch

modeswitch:
Happy to see this move forward. Let me know what you need from me to make this happen.

markandrus:
@modeswitch I think we need to just
1. bump the package.json version to 0.0.60
2. create the v0.0.60 git tag (that should kick off the publishing of the binary)
3. and then publish to NPM

I'm happy to tackle all three, but I'm not a collaborator on the NPM project so I can't publish. I don't think I can add myself, so would you add me?

markandrus:
@caseywebdev I've just published v0.0.60. can you let me know if it's working for you? still working on getting binaries uploaded.

caseywebdev:
Works great! The package built from source since the binaries aren't up like you mentioned but the wall of deprecation logs is gone 👍

PR #272, opened 2016-05-28
Build WebRTC separately

markandrus:
@modeswitch this proof-of-concept separates out the task of building WebRTC from node-webrtc itself. I think this has a few benefits:
1. **Separation of concerns:** Many issues in the issue tracker stem from difficulty building WebRTC; if we could redirect efforts building WebRTC and making binaries available to a separate project, it might refocus our issue tracker.
2. **Easier to keep up-to-date with WebRTC:** AFAIK, [libwebrtc](https://github.com/js-platform/libwebrtc) was set up to simplify/speed up WebRTC builds, but I think it actually binds more than aids our project because we don't get new features, improved Windows support, etc.
3. **Quicker CI build times (e.g. with Travis CI, Appveyor, etc.):** If we can get a separate project building and publishing WebRTC, downloads of WebRTC libraries and headers are just ~~~60 mb~~ (actually, I was building thin libs—will need to adjust this number!); this makes it much easier for us to test OS X, Linux, Windows, etc.
   - It also means we could spin up VMs every once in a while to do the big task of building and publishing WebRTC binaries, and then Travis and Appveyor could be used to build and publish node-webrtc binaries much more easily.

In this particular PR, I've updated node-webrtc to use WebRTC 50, built by my [build-webrtc](https://github.com/markandrus/build-webrtc) project ~~(I've also experimented with a Jakefile, this is still very much WIP)~~.

Some things I think need answered/addressed in this PR:
1. [x] Should probably have a way to build WebRTC from source rather than download a binary (or as a fallback when binaries are unavailable). Concerned about fragility of build scripts. **Answer**: [webrtc-librares-and-headers](https://github.com/markandrus/webrtc-libraries-and-headers) provides a fallback option.
2. [x] ~~Would like to be able to specify WebRTC as a dependency in the package.json, rather than use a script to download and extract it.~~ ~~**Answer:** After looking into this, I think it is not a great idea. The dependency would have to use a postinstall script to pull appropriate binaries for the platform/arch. If we just let node-webrtc pull the binaries, if available, we have one less dependency. May still need a dependency for building WebRTC.~~ Maybe a dependency like [webrtc-librares-and-headers](https://github.com/markandrus/webrtc-libraries-and-headers) does simplify things?
3. [x] What should we do with libwebrtc? Should we combine the work from build-webrtc into libwebrtc? I kind of like the idea of publishing libwebrtc to NPM, e.g. with node-pre-gyp, and specifying it as a dependency of node-webrtc. **Answer:** Let libwebrtc go away and just start tracking WebRTC.
4. [x] Just noticed: build-webrtc was producing thin libs; need to update.


modeswitch:
The `libwebrtc` project was a bit of an experiment to make a stripped-down set of node bindings that would support only data channels. I would consider it a failure for the reasons you gave above. I like the idea of distributing binary versions of the upstream library for common platforms (including raspberry pi), with a fallback option in cases where it needs to be built.

`node-webrtc` as it is now should build fine against the upstream version of libjingle from when libwebrtc was forked. I doubt it would be hard to get it building against headrev.

Personally, I'd be happy with libwebrtc going away in favour of something more maintainable.

markandrus:
> I like the idea of distributing binary versions of the upstream library for common platforms (including raspberry pi), with a fallback option in cases where it needs to be built.

Cool, I can get Linux, OS X, and Windows x64 builds available to start. Raspberry PI will be more challenging. I'll have to read up on it. I'd have to also investigate 32-bit builds.

@modeswitch do you think that fallback to building from source should be the default? Building WebRTC from source is such a resource- and time-intensive process (especially compared to libwebrtc) that I wonder if it should be opt-in, e.g. via an environment variable like `BUILD_WEBRTC=1`.

markandrus:
I've got this branch building on Linux and OS X with Node 4–6 using Travis CI. Trying to get Appveyor working for Windows builds. **EDIT 1:** [Appveyor is working now, too](https://ci.appveyor.com/project/markandrus/node-webrtc), ~~but there are some bugs that keep Node running even after all PeerConnections are closed.~~ **EDIT 2:** There was an issue with the PeerConnectionFactory not getting destructed—this is fixed and now Appveyor builds are green.

markandrus:
@modeswitch now that the proof-of-concept is working with Node 4–6 on Linux, OS X, and Windows via Travis CI and Appveyor, what do you think about the overall design?
- node-webrtc attempts to download a pre-compiled binary using node-pre-gyp.
- If unavailable, node-webrtc falls back to building from source.
  - `binding.gyp` triggers an install of the [webrtc-libraries-and-headers](https://github.com/markandrus/webrtc-libraries-and-headers) package.
  - webrtc-libraries-and-headers attempts to download pre-compiled libraries and header files.
  - If unavailable, webrtc-libraries-and-headers installs [build-webrtc](https://github.com/markandrus/build-webrtc) and attempts to build from source.

I haven't published webrtc-libraries-and-headers or build-webrtc to NPM yet. I'd like to get your feedback on whether this is a good approach or not. I'm also thinking that webrtc-libraries-and-headers should be versioned according to which branch-head it packages, e.g.
- webrtc-libraries-and-headers 0.50.\* — branch-heads/50
- webrtc-libraries-and-headers 0.51.\* — branch-heads/51
- webrtc-libraries-and-headers 0.52.\* — branch-heads/52
- etc.

What do you think?

Globik:
Nice job, @markandrus

JefStat:
> Raspberry PI will be more challenging. I'll have to read up on it. I'd have to also investigate 32-bit builds.

I'm definitely interested in seeing a 32bit binary for debian.

modeswitch:
> @modeswitch do you think that fallback to building from source should be the default? Building WebRTC from source is such a resource- and time-intensive process (especially compared to libwebrtc) that I wonder if it should be opt-in, e.g. via an environment variable like BUILD_WEBRTC=1.

Building from source is practically opt-in anyway, since you must configure a compiler and a bunch of library dependencies. If we use `BUILD_WEBRTC=1` then the build process can emit a sane error message in cases where the build system isn't configured instead of failing part way through the build process.

> @modeswitch now that the proof-of-concept is working with Node 4–6 on Linux, OS X, and Windows via Travis CI and Appveyor, what do you think about the overall design?

If you put the branch and the commit, we can let node-webrtc specify what it needs to build against and either pull the correct libraries-and-headers or pull the correct commit to build from source. Otherwise this looks like a good approach.

markandrus:
> Building from source is practically opt-in anyway, since you must configure a compiler and a bunch of library dependencies. If we use `BUILD_WEBRTC=1` then the build process can emit a sane error message in cases where the build system isn't configured instead of failing part way through the build process.

OK cool, so let's use this.

> If you put the branch and the commit, we can let node-webrtc specify what it needs to build against and either pull the correct libraries-and-headers or pull the correct commit to build from source. Otherwise this looks like a good approach.

OK, so you'd propose foregoing a webrtc-libraries-and-headers NPM package and instead folding [the download code](https://github.com/markandrus/webrtc-libraries-and-headers/blob/fc0c2a69e13b5aecd9158d6f6285aa9c34c307ee/index.js#L39) from webrtc-libraries-and-headers into node-webrtc (we'd also more clearly specify the WebRTC branch-head and commit so that either the binaries can be pulled or the appropriate commit can be pulled and built from source). I can get behind this. **EDIT:** I've implemented this in 88818ab.

markandrus:
@modeswitch mind taking one more look at this?

markandrus:
I am going to go ahead and merge this. I think it is a step in the right direction, and I have some upcoming changes that depend on this. If we see any issues we can address them as they arise.

Globik:
For webrtc libraries and headers one need a compiler with c++11 features? My VC++ 2013 complaints every time by trying to compile anything.

markandrus:
@Globik yes, you'll need C++11 features. Please take a look at the [Appveyor builds](https://ci.appveyor.com/project/markandrus/node-webrtc-7bnua). Perhaps you could replicate that environment?

Globik:
@markandrus  excuse moi, for a webrtc.node aka webrtc.dll should be (via linking stage) one lib dependency - libwebrtc.lib/webrtc.lib? Or much libs? For example, boringssl.lib, libwebrtc.lib, ws2_32.lib, ... dummy-webrtcDatachannelAPI.lib...etc?
I looked in to appveyor link, nothing special there are.
I installed msvc 2015 with update 3 and now all features of c++11 are available. Fine. Only my head around libs for webrtc.node. Those unresolved external referencies... Simply by hand trying to compile with my custom binding.gyp.
Yes, I have webrtc lib dependencies (prebuilded by other man nicked Radioman on github), of him downloaded those libs and they are now in a subdirectory waiting for the link.exe.
Here they are:  https://github.com/radioman/WebRtc.NET/tree/master/WebRtc.NET/libd
Should I all those 200 libs to link to the webrtc.node?

Issue #285, opened 2016-08-17
Check that CreatePeerConnectionFactory actually succeeds


markandrus:
CreatePeerConnectionFactory can fail if initializing the WebRtcVoiceEngine fails. I just reproduced this on an Ubuntu VM:

```
(webrtcvoiceengine.cc:529): WebRtcVoiceEngine::WebRtcVoiceEngine
(webrtcvoiceengine.cc:252): WebRtc VoiceEngine codecs:
(webrtcvoiceengine.cc:279): ISAC/16000/1 (103)
(webrtcvoiceengine.cc:279): ISAC/32000/1 (104)
(webrtcvoiceengine.cc:279): PCMU/8000/1 (0)
(webrtcvoiceengine.cc:279): PCMA/8000/1 (8)
(webrtcvoiceengine.cc:303): Unexpected codec: PCMU/8000/2 (110)
(webrtcvoiceengine.cc:303): Unexpected codec: PCMA/8000/2 (118)
(webrtcvoiceengine.cc:279): ILBC/8000/1 (102)
(webrtcvoiceengine.cc:279): G722/8000/1 (9)
(webrtcvoiceengine.cc:303): Unexpected codec: G722/8000/2 (119)
(webrtcvoiceengine.cc:279): opus/48000/2 (111)
(webrtcvoiceengine.cc:279): CN/8000/1 (13)
(webrtcvoiceengine.cc:279): CN/16000/1 (105)
(webrtcvoiceengine.cc:279): CN/32000/1 (106)
(webrtcvoiceengine.cc:279): telephone-event/8000/1 (126)
(webrtcvoiceengine.cc:279): red/8000/1 (127)
(webrtcvideoengine2.cc:496): WebRtcVideoEngine2::WebRtcVideoEngine2()
(webrtcvoiceengine.cc:556): WebRtcVoiceEngine::Init
(webrtcvoiceengine.cc:571): VoiceEngine 4.1.0
(webrtcvoiceengine.cc:1013): webrtc: CheckPlatform
(webrtcvoiceengine.cc:1013): webrtc: current platform is LINUX
(webrtcvoiceengine.cc:1013): webrtc: CreatePlatformSpecificObjects
(webrtcvoiceengine.cc:1013): webrtc: attempting to use the Linux PulseAudio APIs...
(webrtcvoiceengine.cc:1013): webrtc: Can't load libpulse.so.0 : libpulse.so.0: cannot open shared object file: No such file or directory
[...]

PR #287, opened 2016-08-23
Use a "fake" AudioDeviceModule

markandrus:
This is just one possible approach for improving #285. As mentioned in that issue, on platforms that provide neither PulseAudio nor ALSA, the default AudioDeviceModule will fail. One work around is to use FakeAudioDevice (I've confirmed that this succeeds on the aforementioned platforms). Before merging, I think it makes sense to
- Test with FakeAudioCaptureModule (this was not originally working for me, but I've since updated libwebrtc.a—so maybe it will work again).
- Test with audio MediaStreamTracks (make sure passing "/dev/null" to the FakeAudioDevice isn't an issue)
- Create a plan for choosing between AudioDeviceModule implementations, e.g.

``` js
const wrtc = require('wrtc');
const pc = new wrtc.RTCPeerConnection({
  audioDeviceModule: 'fake',  // or, 'default'
  iceServers: []
});
```
- Still need to do the `nullptr` check on the return value of CreatePeerConnectionFactory.

### DIFF
--- a/src/peerconnection.cc
+++ b/src/peerconnection.cc
@@ -6,6 +6,8 @@
 #include "webrtc/api/mediaconstraintsinterface.h"
 #include "webrtc/api/test/fakeconstraints.h"
 #include "webrtc/base/refcount.h"
+#include "webrtc/system_wrappers/include/clock.h"
+#include "webrtc/test/fake_audio_device.h"
 
 #include "common.h"
 #include "create-answer-observer.h"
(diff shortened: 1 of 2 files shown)


manuguerra:
I tried the changes, and they indeed fix the issue I was seeing on a system that doesn't have alsa (I also had to build libwebrtc using `include_pulse_audio=0` to fix the `XOpenDisplay` issue). I noticed an assert error when running the debug build, but the release version is fine.

The problem that I've noticed is high CPU usage; with FakeAudioDevice, each connection seems to be using about 3 times as much CPU than when using ALSA.

markandrus:
> I tried the changes, and they indeed fix the issue I was seeing on a system that doesn't have alsa

Awesome 👍 

> The problem that I've noticed is high CPU usage; with FakeAudioDevice, each connection seems to be using about 3 times as much CPU than when using ALSA.

Ah, that's unfortunate. We could also try with FakeAudioCaptureModule. It's API is something like

``` js
rtc::scoped_refptr<FakeAudioCaptureModule> facd = FakeAudioCaptureModule::Create();
```

It might be lighter weight. If we still run into these issues I suppose we could always write an AudioDeviceModule that _really_ does nothing...

manuguerra:
I'll try using the FakeAudioCapture later. I also like the idea of writing our own super lightweight AudioDeviceModule, decoupled from the Chromium tests.
In the meantime, I'm sticking to ALSA and pointing LD_LIBRARY to a 'transplanted' libasound.

vikeshkhanna:
I have confirmed that, with this change, `RTCPeerConnection` can be created from within a Docker container. Without this change, the constructor hangs indefinitely at the factory call. Thanks @markandrus!

schuyler1d:
This also fixed building for me on Ubuntu trusty 14.04

markandrus:
I'm going to try to land a version of this as part of #333.

Issue #289, opened 2016-08-31
CPU usage should drop as soon as a connection is closed


manuguerra:
For each new connection, there's an increase in CPU usage; currently, it only drops when the gc is triggered and the the `PeerConnection` destructor is called, which can happen quite a while after the connection is actually closed. For single sporadic connections it's fine, but it can be critical if we happen to have many connections in a short period of time.

As a workaround, I'm currently trying dereferencing the `jinglePeerConnection` and `jinglePeerConnectionFactory` pointers in the `Close` method, which seems to be working fine for me; CPU usage drops as soon as the connection is closed, as expected.
Is it a safe solution? Any suggestions?

[ _tested with Node 4.5 on Ubuntu 14, 32 and 64 bit, and Debian Wheezy 32 bit_ ]

markandrus:
@manuguerra feel free to open a pull request and we can take a look at it. This seems like a reasonable change.

mirague:
This sounds like a great improvement to the library, looking forward to see the PR @manuguerra

melv-n:
This sounds like a great improvement. Could this be related to #278?

manuguerra:
Hi @mirague, I'm not sure, do you have some system stats (specially CPU usage per process) from when you observed your service freezing? 
Even if it's not related, this issue should have an impact in your use case; as you're handling with many simultaneous connections, you'll want to free resources as soon as a connection is closed.

melv-n:
@manuguerra As soon as the freeze happens (after closing a bunch of connections or opening too many) the process can shoot to 100% CPU usage after which the process hangs and then the CPU goes back to a low stable point of < 2% (process stays frozen). Sometimes a SEGFAULT is thrown.

PR #291, opened 2016-09-20
Drop CPU usage as soon as connection is closed

manuguerra:
Currently, the `jinglePeerConnection` and `jinglePeerConnectionFactory` pointers are being nulled only in the destructor; therefore, the threads associated to the peer connection keep running and consuming CPU until the garbage collector kicks in, which can take a while in some cases and can be an issue when handling many connections.

This PR addresses the issue (#289) by also nulling those pointers in the `Close` method. The CPU usage associated with the connection should drop instantly.

The PR also introduces two minor changes to the API:
- `createIceCandidate` will return `undefined` if the connection is already closed when the method is called
- if the connection is already closed when `getStats` is called, the `onFailure` callback will return a `data channel is closed` message.

I've been testing on Ubuntu 14.04 (32 and 64 bit) and on Debian Wheezy (32 bit), node versions 4.5 and 6.5.

### DIFF
--- a/lib/peerconnection.js
+++ b/lib/peerconnection.js
@@ -370,8 +370,12 @@ function RTCPeerConnection(configuration, constraints) {
       args: [label, dataChannelDict]
     });
 
-    dataChannels[label] = channel;
-    return new RTCDataChannel(channel);
+    // channel will be undefined if dataChannel was closed before calling this function
+    // (see peerconnection.cc)
+    if (channel) {  
+        dataChannels[label] = channel;
+        return new RTCDataChannel(channel);
+    }
   };
 
   this.getStats = function getStats(onSuccess, onFailure) {
--- a/src/peerconnection.cc
+++ b/src/peerconnection.cc
@@ -436,6 +436,9 @@ NAN_METHOD(PeerConnection::Close) {
   PeerConnection* self = Nan::ObjectWrap::Unwrap<PeerConnection>(info.This());
   self->_jinglePeerConnection->Close();
 
+  self->_jinglePeerConnection = nullptr;
+  self->_jinglePeerConnectionFactory = nullptr;
+
   TRACE_END;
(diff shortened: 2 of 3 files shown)


markandrus:
Thanks for the PR, @manuguerra. One thing I realized after looking this over: we'll need to guard accesses to `_jinglePeerConnection` after `Close`, e.g. in methods like `CreateOffer` and getters like `GetRemoteDescription`. I'm wondering: do we get the same benefits just null-ing the factory or do we need to null the PeerConnection, too?

manuguerra:
Thanks for taking a look, @markandrus! 

You're right, there a few methods where we should guard accesses to `_jinglePeerConnection`; I'll push some updates later today.

I tried nulling just the factory, but it doesn't solve the issue, it seems we need to null the connection indeed.

markandrus:
Thank you @manuguerra

markandrus:
Hi @feross,

Sorry about the bug. Yes, it does mean that unfortunately. I overlooked that change when I merged. According to http://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-createdatachannel, we should throw an InvalidStateError:

> When the createDataChannel method is invoked, the user agent must run the following steps.
>
> 1. Let connection be the RTCPeerConnection object on which the method is invoked.
> 
> 2. If connection's [[isClosed]] slot is true, throw an InvalidStateError.

I'll try to write a test for this and add the expected behavior this week, although I'm also open to PRs. Should be able to land this update in v0.0.63. In general, we should be following the spec more closely (and hopefully generating more interfaces directly from the WebIDL; I have a side project to build some of this stuff out, but have been busy with other things...).

Best,
Mark

feross:
@markandrus Thanks for the update. I think we're handling this gracefully in `simple-peer` now, so no rush on the change. Just wanted to make sure I raised the point so you were aware. Cheers!

PR #294, opened 2016-10-26
Test for node 7

caseywebdev:
Currently there isn't a build for node 7, so ideally if this passes a new version could be cut with builds uploaded to s3.


caseywebdev:
Looks like appveyor may not support node 7 yet but the travis builds seem to work fine in node 7.

markandrus:
@caseywebdev I re-ran the Appveyor build, and it looks like Node 7 is available now. So the tests are passing. Thanks!

melv-n:
Fantastic, thanks @caseywebdev!

PR #295, opened 2016-11-09
Implemented parsing ICE servers from configuration object

ryskov:
Implemented configurable ICE servers. Now reads and parses ICE server configuration from 'iceServers' key in PeerConnection configuration. 

Implementation might be a bit rough, but should be safe. I am by no means a C++ developer, so feel free to correct/leave comments. 

Anyhow, i think this is better than hardcoding it.


markandrus:
Looks good, @ryskov—thank you.

PR #313, opened 2017-04-07
Make "urls" parameter of "iceServers" spec compliant

### LINKED ISSUE #312: Non-compliant iceServer config
Passing an config in this format will not set the ice servers:

```
{
    iceServers: [
        urls: 'stun:stun.l.google.com:19302'
    ]
}
```
because `wrtc` expects the property `urls` to be an array, and `url` is used for strings. This isn't spec compliant. See: 

https://github.com/js-platform/node-webrtc/blob/develop/src/peerconnection.cc#L297

https://www.w3.org/TR/webrtc/#dom-rtcconfiguration-iceservers
https://www.w3.org/TR/webrtc/#dom-rtciceserver

t-mullen:
Fixes #312

### DIFF
--- a/src/peerconnection.cc
+++ b/src/peerconnection.cc
@@ -288,13 +288,13 @@ NAN_METHOD(PeerConnection::New) {
               Local<Value> iceValue = iceServerObj->Get(iceProps->Get(y)->ToString());
 
               // Handle each field by casting the data and assigning to our iceServer intsance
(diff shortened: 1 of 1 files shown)


feross (approved):
Nice!

markandrus:
Thanks @RationalCoding! 

I'm not quite sure what's up with the OS X Node 5 and 7 builds on Travis CI right now. Will look into them.

PR #331, opened 2017-09-04
add ice candidate parse error description if any

oklas:
When working together with package `react-native-webrtc` error 

> Failed to set ICE candidate

is fired adding this code add some details about error in some case.

However really `jinglePeerConnection` is not null and `sdpParseError.description` is empty but
`self->_jinglePeerConnection->AddIceCandidate(ci)` return false and error is fired.


markandrus:
LGTM

PR #332, opened 2017-09-24
Refactor event handling, refactor data type conversions, and add C++ unit tests

markandrus:
~~No significant changes here. Mainly, I've setup CLion and it recommends a
handful of cosmetic changes to improve the code. I've also tweaked the
AsyncEvent code, because I think we were leaking the `data` members. We can
factor out the AsyncEvent code between PeerConnection and DataChannel.~~

There are a handful of changes I am working on here:

- [x] Updated LICENSE to 2017 and to refer to the "node-webrtc project authors".
- [x] Factored out EventQueue and EventLoop classes
- [x] PeerConnection, DataChannel, and DataChannelObserver now extend these
- [x] Events are handled via double dispatch (couldn't get RTTI/`dynamic_dispatch` enabled in MSVS)
- [x] Added CMake support (in part, using NodeJS.cmake from node-cmake)
- [x] Added some FP-style conversion functions (Converters return Validations for data types, letting us collect errors; syntactic sugar for Map (%) and Apply (*) inspired by [FTL](https://github.com/beark/ftl))
- [x] Added support for more W3C-specced dictionaries, even if we aren't supporting them with our current webrtc.org build (RTCOAuthCredential, additional members in RTCConfiguration, etc.)
- [x] Added C++ unit tests—if node-webrtc is built in Debug mode, you'll be able to run `./bin/test`, which just calls into [Catch](https://github.com/philsquared/Catch)-based unit test. Shouldn't be too hard to hook up Valgrind to these.
- [ ] Add native Promise support
- [ ] Minimize or remove `lib/peerconnection.js`


markandrus:
Excluding the C++ unit tests, the useful refactoring in this PR (Converter, EventQueue, EventLoop, PromiseFulfillingEventLoop, etc.) have landed in other PRs.

PR #333, opened 2017-11-03
Try building with Node 9

markandrus:
@pvh


markandrus:
With Node 9.x.x, the V8 headers now include `constexpr`s which are not supported by MSVS 2013. We need to go to 2015. Annoyingly, we'll also have to rebuild WebRTC using 2015. Maybe this is a good opportunity to bump the WebRTC version we use, try out GN, etc.

markandrus:
I'm going to see about building with https://github.com/aisouard/libwebrtc.

PR #342, opened 2017-12-24
Fix segfaults in RTCDataChannel and RTCPeerConnection by not freeing objects until destruction

### LINKED ISSUE #236: Datachannel onclose event does not trigger
I do not get the onclose event from datachannels. I've tested this with the latest 4 versions and both on Linux (Ubuntu 14.04) and OS X (Maverick). I've tested with both the example code and custom code but the onclose event is not triggered. Have I missed something here?

### LINKED ISSUE #325: segfault when client disconnects
I detect client disconnects through a websocket and try to stop sending data asap. Either way wrtc should not segfault.

Extract from my own log:

...
sending frame
sending frame
sending frame
sending frame
sending frame
sending frame
client closed
PID 21292 received SIGSEGV for address: 0x0
/home/zubzub/git/greenfield/node_modules/segfault-handler/build/Release/segfault-handler.node(+0x1bb8)[0x7fc436dfbbb8]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x11390)[0x7fc4474da390]
/home/zubzub/git/greenfield/node_modules/wrtc/build/wrtc/v0.0.62/Release/node-v57-linux-x64/wrtc.node(_ZN11node_webrtc11DataChannel13GetReadyStateEN2v85LocalINS1_6StringEEERKN3Nan20PropertyCallbackInfoINS1_5ValueEEE+0x54)[0x7fc4440d7bf4]
/home/zubzub/git/greenfield/node_modules/wrtc/build/wrtc/v0.0.62/Release/node-v57-linux-x64/wrtc.node(+0x7f1c7)[0x7fc4440d71c7]
[0x38ee83d51080]

nazar-pc:
This fixes #236 and fixes #325.

Hopefully, this can be merged and released soon.

P.S. I'm not a C/C++ developer even remotely and might have done something horribly wrong, but it seems to work correctly.

### DIFF
--- a/src/datachannel.cc
+++ b/src/datachannel.cc
@@ -229,37 +229,39 @@ NAN_METHOD(DataChannel::Send) {
 
   DataChannel* self = Nan::ObjectWrap::Unwrap<DataChannel>(info.This());
 
-  if (info[0]->IsString()) {
-    Local<String> str = Local<String>::Cast(info[0]);
-    std::string data = *String::Utf8Value(str);
+  if (self->_jingleDataChannel != nullptr) {
+    if (info[0]->IsString()) {
+      Local<String> str = Local<String>::Cast(info[0]);
+      std::string data = *String::Utf8Value(str);
 
-    webrtc::DataBuffer buffer(data);
-    self->_jingleDataChannel->Send(buffer);
-  } else {
-#if NODE_MINOR_VERSION >= 11 || NODE_MAJOR_VERSION > 0
-    Local<v8::ArrayBuffer> arraybuffer;
-
-    if (info[0]->IsArrayBuffer()) {
-      arraybuffer = Local<v8::ArrayBuffer>::Cast(info[0]);
+      webrtc::DataBuffer buffer(data);
+      self->_jingleDataChannel->Send(buffer);
     } else {
-      Local<v8::ArrayBufferView> view = Local<v8::ArrayBufferView>::Cast(info[0]);
-      arraybuffer = view->Buffer();
-    }
+#if NODE_MINOR_VERSION >= 11 || NODE_MAJOR_VERSION > 0
+      Local<v8::ArrayBuffer> arraybuffer;
 
-    v8::ArrayBuffer::Contents content = arraybuffer->GetContents();
-    rtc::CopyOnWriteBuffer buffer(static_cast<char*>(content.Data()), content.ByteLength());
+      if (info[0]->IsArrayBuffer()) {
+        arraybuffer = Local<v8::ArrayBuffer>::Cast(info[0]);
+      } else {
+        Local<v8::ArrayBufferView> view = Local<v8::ArrayBufferView>::Cast(info[0]);
+        arraybuffer = view->Buffer();
+      }
+
+      v8::ArrayBuffer::Contents content = arraybuffer->GetContents();
+      rtc::CopyOnWriteBuffer buffer(static_cast<char*>(content.Data()), content.ByteLength());
 
 #else
-    Local<Object> arraybuffer = Local<Object>::Cast(info[0]);
-    void* data = arraybuffer->GetIndexedPropertiesExternalArrayData();
-    uint32_t data_len = arraybuffer->GetIndexedPropertiesExternalArrayDataLength();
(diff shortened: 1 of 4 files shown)


markandrus:
Thanks, @nazar-pc! Do you know why we were null-ing out the `_jinglePeerConnection` and related objects? I thought there was a performance reason. Not sure, but `git blame` should show. I’m AFK at the moment, otherwise would check.

nazar-pc:
Interesting, after reading #291 I've reworked PR into simpler one.
BTW, tests from #291 are failing as of current `develop`, but they work properly with this PR.

src/datachannel.cc
```diff
@@ -229,37 +229,39 @@ NAN_METHOD(DataChannel::Send) {
 
   DataChannel* self = Nan::ObjectWrap::Unwrap<DataChannel>(info.This());
 
-  if (info[0]->IsString()) {
-    Local<String> str = Local<String>::Cast(info[0]);
-    std::string data = *String::Utf8Value(str);
+  if (self->_jingleDataChannel != nullptr) {
```
markandrus, line 232: In a future PR, we should actually throw an InvalidStateError. See [here](https://w3c.github.io/webrtc-pc/#dfn-send()).

src/datachannel.cc
```diff
@@ -293,7 +298,13 @@ NAN_GETTER(DataChannel::GetBufferedAmount) {
...
   DataChannel* self = Nan::ObjectWrap::Unwrap<DataChannel>(info.Holder());
 
-  uint64_t buffered_amount = self->_jingleDataChannel->buffered_amount();
+  uint64_t buffered_amount;
+
+  if (self->_jingleDataChannel != nullptr) {
+    buffered_amount = self->_jingleDataChannel->buffered_amount();
+  } else {
+    buffered_amount = 0;
+  }
```
markandrus: WDYT about writing this as

```c++
uint64_t buffered_amount = self->_jingleDataChannel
  ? self->_jingleDataChannel->buffered_amount()
  : 0;
```

Also, in the future, this property should behave differently. See [here](http://w3c.github.io/webrtc-pc/#dom-rtcdatachannel-bufferedamount):

> the attribute does not reset to zero once the channel closes

That being said, the point of this PR is just to fix a segfault, not enforce spec-compliance, so we should punt this to a future PR.

src/datachannel.cc
```diff
@@ -304,7 +315,13 @@ NAN_GETTER(DataChannel::GetLabel) {
...
   DataChannel* self = Nan::ObjectWrap::Unwrap<DataChannel>(info.Holder());
 
-  std::string label = self->_jingleDataChannel->label();
+  std::string label;
+
+  if (self->_jingleDataChannel != nullptr) {
+    label = self->_jingleDataChannel->label();
+  } else {
+    label = "";
+  }
```
markandrus: Ditto re: ternary as above.

src/datachannel.cc
```diff
@@ -315,7 +332,13 @@ NAN_GETTER(DataChannel::GetReadyState) {
...
   DataChannel* self = Nan::ObjectWrap::Unwrap<DataChannel>(info.Holder());
 
-  webrtc::DataChannelInterface::DataState state = self->_jingleDataChannel->state();
+  webrtc::DataChannelInterface::DataState state;
+
+  if (self->_jingleDataChannel != nullptr) {
+    state = self->_jingleDataChannel->state();
+  } else {
+    state = webrtc::DataChannelInterface::kClosed;
+  }
```
markandrus: Ditto re: ternary as above.

src/datachannel.cc
markandrus: https://github.com/js-platform/node-webrtc/issues/343

markandrus:
I left a few comments. Please let me know what you think; otherwise, the changes look good to me. Please also add yourself to the `AUTHORS` file if you’d like.

nazar-pc:
Refactored as suggested, comments make perfect sense to me.

markandrus:
Thanks! @nazar-pc will try to get a release out in next few days

3mcd:
Great work!

PR #352, opened 2018-01-19
Destroy mutexes in destructor

nazar-pc:
I have no idea how to use libuv, but from usage examples I found it seems that `uv_mutex_destroy` is needed.


markandrus:
LGTM, thank you

Issue #351, opened 2018-01-18
High CPU load because of C++ code


nazar-pc:
In my app I found that CPU usage is extremely high, even for 2 nodes communicating with each other (and creating spare half-open connections in the mean time). JavaScript side was not an issue, it had very low load and I decided to profile it, here is what I've got:
<details>

```
nazar-pc@nazar-pc /w/g/detox-core> node --prof-process isolate-0x24ac240-v8.log 
line 3806: unknown code state: 4}:([0-9a-fA-F]){1\
Statistical profiling result from isolate-0x24ac240-v8.log, (818 ticks, 85 unaccounted, 0 excluded).

 [Shared libraries]:
   ticks  total  nonlib   name
    121   14.8%          /usr/bin/node
      6    0.7%          /lib/x86_64-linux-gnu/libc-2.26.so
      1    0.1%          [vdso]
      1    0.1%          /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.24
      1    0.1%          /lib/x86_64-linux-gnu/ld-2.26.so

 [JavaScript]:
   ticks  total  nonlib   name
      4    0.5%    0.6%  LazyCompile: ~deflateSetDictionary /web/github/detox-core/node_modules/pako/lib/zlib/deflate.js:1773:30
      4    0.5%    0.6%  Builtin: InterpreterEntryTrampoline
      2    0.2%    0.3%  Stub: StoreFastElementStub
      2    0.2%    0.3%  StoreIC: A store IC from the snapshot
      2    0.2%    0.3%  LazyCompile: ~send_bits /web/github/detox-core/node_modules/pako/lib/zlib/trees.js:213:19
      2    0.2%    0.3%  LazyCompile: ~longest_match /web/github/detox-core/node_modules/pako/lib/zlib/deflate.js:231:23
      2    0.2%    0.3%  LazyCompile: ~inflate_fast /web/github/detox-core/node_modules/pako/lib/zlib/inffast.js:61:39
      2    0.2%    0.3%  LazyCompile: ~B /web/github/detox-core/node_modules/jssha/src/sha3.js:26:123
      2    0.2%    0.3%  KeyedLoadIC: A keyed load IC from the snapshot
      2    0.2%    0.3%  Function: WASM_function#20:module:<?> (wasm):0:0
      2    0.2%    0.3%  Builtin: FastNewObject
      2    0.2%    0.3%  Builtin: Call_ReceiverIsAny
[...]

PR #353, opened 2018-02-03
Improve fake audio device

### LINKED ISSUE #340: Exceedingly high CPU usage on Windows
OS: Windows 10 (1703).

When version 0.0.63 is used, the percentage of CPU usage by a single node process running wrtc remains at a level of 30% to 45% (CPU: Intel i5, 4 cores). This is not the case when 0.0.62 is used. The CPU usage remains quite low even on a NCU (next computing unit with low end Intel Atom CPU)

It looks normal on Linux systems, even for virtual machines ...

[Update]:
It seems to yield well to other processes when they need to use the CPU. But when the system is otherwise idle, its CPU usage level returns to high. This could generates potential heat problems on small devices ...

markandrus:
I think this should fix #340 and #351. I ported (and re-tabbed) the FakeAudioDevice from WebRTC 9a0a17fb7a4a594825d7fb3c676e58476e892cfc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+0.0.66 (in progress)
+====================
+
+Bug Fixes
+---------
+
+- The FakeAudioDeviceModule introduced in v0.0.63 used too much CPU. v0.0.66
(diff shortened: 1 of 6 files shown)


markandrus:
Eh, still some issues here.

PR #357, opened 2018-02-08
Fix CPU regression

markandrus:


nazar-pc:
Looks very promising, tested `test.js` from #340, the first (`develop`) have failed and second (`peerconnectionfactory`) succeeded reasonably quickly:
```
nazar-pc@nazar-pc /w/g/node-webrtc> bash run-test.sh 
pure virtual method called
terminate called without an active exception
run-test.sh: рядок 5: 28249 Перервано      (збережено знімок оперативної пам’яті) node test.js > /dev/null
654.019 1067

real    0m17,363s
user    0m58,417s
sys     2m1,607s
nazar-pc@nazar-pc /w/g/node-webrtc> bash run-test.sh 
79.2508  142

real    0m5,194s
user    0m4,301s
sys     0m5,404s
```

Awesome!

src/datachannel.cc
```diff
@@ -220,6 +218,10 @@ void DataChannel::OnStateChange() {
   TRACE_CALL;
   StateEvent* data = new StateEvent(_jingleDataChannel->state());
   QueueEvent(DataChannel::STATE, static_cast<void*>(data));
+  if (_jingleDataChannel->state() == webrtc::DataChannelInterface::kClosed) {
+    _jingleDataChannel->UnregisterObserver();
+    _jingleDataChannel = nullptr;
+  }
```
markandrus, line 224: Note to self: needed to move this out of `Run`, because, AFAICT, after `OnStateChange` is invoked with state `kClosed`, the underlying DataChannel will be freed. Later, when we read `state` from the DataChannel in `Run`, we were getting segfaults.

nazar-pc:
Wow, CPU usage using test from #351 went from 10 cores fully loaded to just over 1 core!

src/peerconnection.cc
```diff
@@ -53,15 +50,11 @@ PeerConnection::PeerConnection(webrtc::PeerConnectionInterface::IceServers iceSe
...
   configuration.servers = iceServerList;
 
-  webrtc::FakeConstraints constraints;
-  constraints.AddOptional(webrtc::MediaConstraintsInterface::kEnableDtlsSrtp, webrtc::MediaConstraintsInterface::kValueTrue);
-  // FIXME: crashes without these constraints, why?
-  constraints.AddMandatory(webrtc::MediaConstraintsInterface::kOfferToReceiveAudio, webrtc::MediaConstraintsInterface::kValueFalse);
-  constraints.AddMandatory(webrtc::MediaConstraintsInterface::kOfferToReceiveVideo, webrtc::MediaConstraintsInterface::kValueFalse);
+  // TODO(mroberts): Read `factory` (non-standard) from RTCConfiguration?
+  _factory = PeerConnectionFactory::GetOrCreateDefault();
+  _shouldReleaseFactory = true;
```
markandrus, line 55: In the future, it might make sense to allow

```js
const { RTCPeerConnection, RTCPeerConnectionFactory } = require('wrtc')

const factory = new RTCPeerConnectionFactory({
  audioLayer: 'dummy' // or any of the other AudioLayers
})

const pc = new RTCPeerConnection({
  factory // non-standard addition to RTCConfiguration
})
```

This would allow experimenting with different AudioDeviceModules, as well as control over threads (since there is one signaling and one worker thread per RTCPeerConnectionFactory). Not sure if more threads == better, but it would be an option. Also, for reference, the AudioLayers are

```cpp
enum AudioLayer {
  kPlatformDefaultAudio = 0,
  kWindowsWaveAudio = 1,
  kWindowsCoreAudio = 2,
  kLinuxAlsaAudio = 3,
  kLinuxPulseAudio = 4,
  kAndroidJavaAudio = 5,
  kAndroidOpenSLESAudio = 6,
  kAndroidJavaInputAndOpenSLESOutputAudio = 7,
  kDummyAudio = 8
};
```

markandrus:
Thanks for testing, @nazar-pc! Will release this evening

PR #359, opened 2018-02-10
Ref() and Unref() ObjectWrap instances w/ EventLoops

markandrus:
@nazar-pc please take a look. I was seeing segfaults when accessing `self->handle()` in `Run`. Both node_webrtc::PeerConnection and node_webrtc::DataChannel run an event loop (see `Run`). I think the issue is that these objects could be freed despite the fact that `Run` was still running and trying to access the instances via `async`. [Docs for nan::ObjectWrap](https://github.com/nodejs/nan/blob/master/doc/object_wrappers.md) mention `Ref()` and `Unref()` methods to be used if an object is part of an event loop. So the new pattern here is to `Ref()` in `New` and `Unref()` at the end of `Run` when `do_shutdown` is `true`.

Edit: I ran a tweaked version locally where I logged in the PeerConnection and DataChannel destructors; I see PeerConnections destruct, but rarely DataChannels, so there may be a bug causing DataChannels to leak. Should address before merge...

Edit 2: I believe the DataChannels aren't freed because they are retained in an object inside `lib/peerconnection.js`. Removing this object (I cannot see why it is necessary) allows them to be freed.

### DIFF
--- a/lib/peerconnection.js
+++ b/lib/peerconnection.js
@@ -18,7 +18,6 @@ function RTCPeerConnection(configuration, constraints) {
   var remoteType = null;
   var queue = [];
   var pending = null;
-  var dataChannels = {};  // open data channels, indexed by label
 
   EventTarget.call(this);
 
@@ -85,17 +84,7 @@ function RTCPeerConnection(configuration, constraints) {
     self.dispatchEvent(new RTCPeerConnectionIceEvent('icecandidate', { candidate: icecandidate }));
   };
 
-  pc.onsignalingstatechange = function onsignalingstatechange(state) {
-    var stateString = self.RTCSignalingStates[state];
-
-    if ('closed' === stateString) {
-      Object.keys(dataChannels).forEach(function(label) {
-        dataChannels[label].shutdown();
-
-        delete dataChannels[label];
-      });
-    }
-
+  pc.onsignalingstatechange = function onsignalingstatechange() {
     self.dispatchEvent({ type: 'signalingstatechange' });
   };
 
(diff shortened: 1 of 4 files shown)


nazar-pc:
Got only 2 segfaults in a few dozens of runs locally, at least second time it happened at the end after destruction of all of the opened connections.

Currently trying to get segfault under gdb to extract a useful stacktrace, but it certainly works much better now.

nazar-pc:
Got a segfault in the process of destroying of connections (not after), here is backtrace from gdb:
```
Thread 1 "node" received signal SIGSEGV, Segmentation fault.
0x000000000294b498 in ?? ()
(gdb) backtrace
#0  0x000000000294b498 in ?? ()
#1  0x00007fffdb469dbd in webrtc::internal::SynchronousMethodCall::Invoke(rtc::Location const&, rtc::Thread*) () from /web/github/node-webrtc/build/Release/wrtc.node
#2  0x00007fffdb49a950 in webrtc::DataChannelProxyWithInternal<webrtc::DataChannelInterface>::Close() () from /web/github/node-webrtc/build/Release/wrtc.node
#3  0x00007fffdb45cc32 in Nan::imp::FunctionCallbackWrapper () from /web/github/node-webrtc/build/Release/wrtc.node
#4  0x000038a8afb36d87 in ?? ()
#5  0x00007fffffffc7f8 in ?? ()
#6  0x00007fffffffc830 in ?? ()
#7  0x0000000000000000 in ?? ()
```

Not sure why `??`, can I build it in some other way to give you more info?

markandrus:
Thanks! That’s a helpful backtrace—so maybe closing in a callback—like statechange—might do something weird?

nazar-pc:
I think this can be merged and mentioned segfault fixed separately if it is not caused by these new changes.

PR #361, opened 2018-02-17
Fix segfault on datachannel close

### LINKED ISSUE #360: Closing DataChannel after PeerConnection can segfault
@nazar-pc I think I reproduced your segfault from #359. The following

```js
'use strict';

const { RTCPeerConnection } = require('.');

const dc = (() => {
  const pc = new RTCPeerConnection();
  const dc = pc.createDataChannel();
  pc.close();
  return dc;
})();

dc.close();
```

yields

```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x38)
  * frame #0: 0x0000000104c6678f wrtc.node`webrtc::internal::SynchronousMethodCall::Invoke(this=0x00007fff5fbfdcd0, posted_from=0x00007fff5fbfdc58, t=0x0000000101f1dc70) at proxy.h:115 [opt]
    frame #1: 0x0000000104c60a97 wrtc.node`webrtc::DataChannelProxyWithInternal<webrtc::DataChannelInterface>::Close() [inlined] webrtc::MethodCall0<webrtc::DataChannelInterface, void>::Marshal(this=<unavailable>, posted_from=0x00007fff5fbfdc58, t=<unavailable>) at proxy.h:136 [opt]
    frame #2: 0x0000000104c60a52 wrtc.node`webrtc::DataChannelProxyWithInternal<webrtc::DataChannelInterface>::Close(this=0x000000010319e4d0) at datachannel.h:291 [opt]
    frame #3: 0x0000000104806fd8 wrtc.node`node_webrtc::DataChannel::Close(info=0x00007fff5fbfdd50) at datachannel.cc:293
    frame #4: 0x000000010480b7d9 wrtc.node`Nan::imp::FunctionCallbackWrapper(info=0x00007fff5fbfddf0) at nan_callbacks_12_inl.h:174
    frame #5: 0x00000001000c1a99 node`v8::internal::FunctionCallbackArguments::Call(void (*)(v8::FunctionCallbackInfo<v8::Value> const&)) + 425
    frame #6: 0x0000000100122e2b node`v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<false>(v8::internal::Isolate*, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::FunctionTemplateInfo>, v8::internal::Handle<v8::internal::Object>, v8::internal::BuiltinArguments) + 843
    frame #7: 0x00000001001223d1 node`v8::internal::Builtin_Impl_HandleApiCall(v8::internal::BuiltinArguments, v8::internal::Isolate*) + 273
```

markandrus:
@nazar-pc this should fix #360. The issue is that `node_webrtc::PeerConnection` was the only thing holding a reference to `node_webrtc::PeerConnectionFactory`. Once the `RTCPeerConnection` got garbage collected, `node_webrtc::PeerConnection::~PeerConnection` gets invoked, `Release`ing the `node_webrtc::PeerConnectionFactory`. This `Stop`s and frees the `rtc::Thread`s in use, causing the segfault when the `RTCDataChannel` is `close`d.

The solution is to thread the `node_webrtc::PeerConnectionFactory` along to child objects like `node_webrtc::DataChannel` (and any eventual `node_webrtc::MediaStreamTrack`s we might add—cc @thedracle). Then, `node_webrtc::PeerConnection` can still `Release` the `node_webrtc::PeerConnectionFactory`, but it won't get garbage collected until the objects that depend on it are garbage collected.

There were some conditions I tested manually with the code with `TRACING` defined (namely, that destructors for `node_webrtc::PeerConnectionFactory` and `node_webrtc::DataChannel` did fire, and in the right order). We want to make sure these never leak, so we should consider landing some unit tests in the future that check the `TRACING` output to ensure they fire (perhaps also using Node's `--expose-gc` flag and the `gc` function). Until then, I've added the previously failing test, `test/closing-data-channel.js`.

### DIFF
--- a/src/datachannel.cc
+++ b/src/datachannel.cc
@@ -29,15 +29,18 @@ Nan::Persistent<Function> DataChannel::constructor;
 Nan::Persistent<Function> DataChannel::ArrayBufferConstructor;
 #endif
 
-DataChannelObserver::DataChannelObserver(rtc::scoped_refptr<webrtc::DataChannelInterface> jingleDataChannel) {
+DataChannelObserver::DataChannelObserver(std::shared_ptr<node_webrtc::PeerConnectionFactory> factory,
+                                         rtc::scoped_refptr<webrtc::DataChannelInterface> jingleDataChannel) {
   TRACE_CALL;
   uv_mutex_init(&lock);
+  _factory = factory;
   _jingleDataChannel = jingleDataChannel;
   _jingleDataChannel->RegisterObserver(this);
   TRACE_END;
 }
 
 DataChannelObserver::~DataChannelObserver() {
+  _factory = nullptr;
   _jingleDataChannel = nullptr;
   uv_mutex_destroy(&lock);
 }
@@ -73,6 +76,7 @@ DataChannel::DataChannel(node_webrtc::DataChannelObserver* observer)
   uv_mutex_init(&lock);
   uv_async_init(loop, &async, reinterpret_cast<uv_async_cb>(Run));
 
+  _factory = observer->_factory;
   _jingleDataChannel = observer->_jingleDataChannel;
   _jingleDataChannel->RegisterObserver(this);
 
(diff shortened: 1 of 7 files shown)


nazar-pc:
Tested a few scenarios locally and it looks very promising so far 👍

PR #364, opened 2018-02-19
Refactor events

markandrus:
I experimented with this approach in #332 and have extracted it out. IMO, it should make events easier to manage/add to in the future. It also encapsulates the libuv-related code in EventLoop. I want to do some more testing before merging to make sure none of the regressions addressed in recent PRs are re-introduced. I also want to make sure destructors are still firing as expected.


markandrus:
Followup on #387

PR #368, opened 2018-03-13
Port Restriction Policy

### LINKED ISSUE #354: Port Restriction Policy
For my use-case I'd like to be able to restrict the range of ports that the ICE gatherer is able to generate local candidates from - I've seen a policy for this in Chrome (Google.Policies.Chrome::WebRtcUdpPortRange).

Would it be possible to add this feature?
I'm not exactly sure where the binaries come from / what is possible via the bindings...

I'd appreciate it if someone could point me in the right direction, I'm willing to implement this myself.

Thank you,
Nicholas

njohnst:
Adds a custom option 'portRange' to RTCConfiguration in the RTCPeerConnection constructor.  This option takes a string of the form 'MIN_PORT-MAX_PORT'  and restricts the range of ports that the ICE gatherer can draw from (for use i.e. in corporate networks or in port-forwarded client-server applications (see limitation below)).  

Limitation: if a port range contains N ports, then only N peers can connect via UDP, as the native webrtc implementation binds a port for each peer connection (note that an additional N peers may connect via TCP).

Note: the native implementation of PeerConnectionFactory creates an extra instance of rtc::BasicNetworkManager and rtc::BasicPacketSocketFactory that is privately scoped and can't be avoided.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -13,3 +13,4 @@ Manu Raghavan <manu@voxology.co>
 Ross Kukulinski <ross@getyodlr.com>
 Jeppe Ryskov Larsen <jeppe@bjergsted.dk>
 Nazar Mokrynskyi <nazar@mokrynskyi.com>
+Nicholas Johnston <nickjoh@protonmail.ch>
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,7636 @@
+{
+  "name": "wrtc",
+  "version": "0.0.68-dev",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@browserify/acorn5-object-spread": {
+      "version": "5.0.1",
+      "resolved": "https://registry.npmjs.org/@browserify/acorn5-object-spread/-/acorn5-object-spread-5.0.1.tgz",
+      "integrity": "sha512-sFCUPzgeEjdq3rinwy4TFXtak2YZdhqpj6MdNusxkdTFr9TXAUEYK4YQSamR8Joqt/yii1drgl5hk8q/AtJDKA==",
+      "dev": true,
+      "requires": {
+        "acorn": "5.4.1"
+      }
+    },
+    "JSONStream": {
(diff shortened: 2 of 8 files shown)


markandrus:
Thanks so much, @njohnst!

> This option takes a string of the form 'MIN_PORT-MAX_PORT'

WDYT about letting the value of `portRange` be a [LongRange](https://developer.mozilla.org/en-US/docs/Web/API/LongRange) instead of the string?--**EDIT:** actually, the port range isn't long, though, so we'd have to consider it an "UnsignedShortRange", which I haven't seen in any specs, but it should "make sense".--If we specced this as a custom addition to WebRTC's RTCConfiguration Web IDL, it could look something like this

```webidl
partial dictionary RTCConfiguration {
  LongRange portRange;
}
```

and then in JavaScript we can pass

```js
{
  portRange: { min: 10000, max: 20000 }
}
```

Also, are the defaults 0 and 65535? i.e., is specifying

```js
{
  portRange: { min: 0, max: 65535 }
}
```

the same as specifying nothing at all? Or the same as the behavior before this change?

markandrus:
If you're OK with the suggestion above, I'd like to work it in after #370, since I think that will make dictionary parsing more manageable going forward.

njohnst:
Sounds good, I'll take a look at #371 

Regarding the port ranges: the default behavior is minPort = 0, maxPort = 0, this is the same as passing 0-65535.  When I tried to pass 0-10000 it actually allocated a port >50000, so it seems that
passing minPort = 0 causes the port allocator to select any port, regardless of maxPort.

If minPort = 0: OS is asked to give a random port
If minPort > 0: PortAllocator attempts to allocate the first available port >= minPort

N.B. in the case that default values are passed (0-0), the behavior is actually slightly different from before: PortAllocator is always manually passed (even with default values).  In the previous version the PortAllocator is created in the PeerConnectionFactory (using the privately scoped rtc::BasicNetworkManager and rtc::BasicPacketSocketFactory that I mentioned above).

PR #370, opened 2018-03-17
Conversions

markandrus:
Extracted from #332. Idea is to make converting to/from C++ more declarative and extensible going forward by borrowing some functional programming concepts. For example, a relatively complicated dictionary, like RTCConfiguration, can be handled as

```cpp
Validation<RTCConfiguration> Converter<Local<Value>, RTCConfiguration>::Convert(const Local<Value> value) {
  return From<Local<Object>>(value).FlatMap<RTCConfiguration>(
      [](const Local<Object> object) {
        return curry(CreateRTCConfiguration)
            % GetOptional<std::vector<IceServer>>(object, "iceServers", std::vector<IceServer>())
            * GetOptional<IceTransportsType>(object, "iceTransportPolicy", IceTransportsType::kAll)
            * GetOptional<BundlePolicy>(object, "bundlePolicy", BundlePolicy::kBundlePolicyBalanced)
            * GetOptional<RtcpMuxPolicy>(object, "rtcpMuxPolicy", RtcpMuxPolicy::kRtcpMuxPolicyRequire)
            * GetOptional<std::string>(object, "peerIdentity")
            * GetOptional<std::vector<Local<Object>>>(object, "certificates")
            // TODO(mroberts): Implement EnforceRange and change to uint8_t.
            * GetOptional<uint32_t>(object, "iceCandidatePoolSize", 0);
      });
}
```

### DIFF
--- a/examples/bridge.js
+++ b/examples/bridge.js
@@ -136,7 +136,7 @@ wss.on('connection', function(ws) {
       remoteReceived = false;
 
       pc = new webrtc.RTCPeerConnection({
-        iceServers: [{ url: 'stun:stun.l.google.com:19302' }]
+        iceServers: [{ urls: 'stun:stun.l.google.com:19302' }]
       }, {
         optional: [{ DtlsSrtpKeyAgreement: false }]
       });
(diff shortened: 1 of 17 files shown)


markandrus:
There are some additional converters that can be implemented for enums like SignalingState, IceConnectionState, etc., and for RTCSessionDescription, but we can implement those later. Landing this now should make #368 easier.

PR #371, opened 2018-03-17
Port Restriction Policy 2

markandrus:
@njohnst I took #368 and ran with the `min`/`max` idea, building on work from #370. I still need to do some more testing, but ExtendedRTCConfiguration should give us a pattern for adding future custom RTCConfiguration options if we deem them useful. **EDIT:** Also, there are some lint errors to address in the custom-settings.js test.


src/peerconnection.cc
```diff
@@ -64,7 +65,20 @@ PeerConnection::PeerConnection(RTCConfiguration configuration)
...
   _shouldReleaseFactory = true;
 
-  _jinglePeerConnection = _factory->factory()->CreatePeerConnection(configuration, nullptr, nullptr, nullptr, this);
+  auto portAllocator = std::unique_ptr<cricket::PortAllocator>(new cricket::BasicPortAllocator(
+              _factory->getNetworkManager(),
+              _factory->getSocketFactory()
+          ));
+  portAllocator->SetPortRange(
+      configuration.portRange.min.FromMaybe(0),
+      configuration.portRange.max.FromMaybe(65535));
```
njohnst: bracket on separate line ?

njohnst:
Looks great, a lot cleaner to deal with the config parsing this way
Thank you for fixing the lint errors as well
In peerconnection.cc I spotted a few places where the parentheses/line breaks were slightly inconsistent (lines 68-81), not sure if that's worth fixing

src/peerconnection.cc
markandrus: Ah, good catch. I moved the bracket to the same line instead.

markandrus:
Thanks, @njohnst!

PR #377, opened 2018-03-25
Fix "Assertion `handle->flags & UV_CLOSING' failed."

### LINKED ISSUE #376: Assertion failure in Node.js 9.9.0
I've recently upgraded to Node.js 9.9.0 and started occasionally getting this:
```
node: ../deps/uv/src/unix/core.c:250: uv__finish_close: Assertion `handle->flags & UV_CLOSING' failed.

Thread 1 "node" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) backtrace
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff6ab7801 in __GI_abort () at abort.c:79
#2  0x00007ffff6aa739a in __assert_fail_base (fmt=0x7ffff6c2e7d8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x1761717 "handle->flags & UV_CLOSING", file=file@entry=0x17616f5 "../deps/uv/src/unix/core.c", line=line@entry=250, function=function@entry=0x17619b0 "uv__finish_close") at assert.c:92
#3  0x00007ffff6aa7412 in __GI___assert_fail (assertion=0x1761717 "handle->flags & UV_CLOSING", file=0x17616f5 "../deps/uv/src/unix/core.c", line=250, function=0x17619b0 "uv__finish_close") at assert.c:101
#4  0x00000000009bf2b0 in uv_run ()
#5  0x00000000008dbb9d in node::Start(uv_loop_s*, int, char const* const*, int, char const* const*) ()
#6  0x00000000008d7ef3 in node::Start(int, char**) ()
#7  0x00007ffff6a98b97 in __libc_start_main (main=0x8a1e60 <main>, argc=2, argv=0x7fffffffe458, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffe448) at ../csu/libc-start.c:310
#8  0x00000000008a1f51 in _start ()
```

Can be reproduced by running tests from https://github.com/Detox/core several times.

I'm not 100% positive it is caused by `node-webrtc`, but is is a single native extension I have and it is not possible to run tests without it, so I thought it might be worth to let you know.

markandrus:
@nazar-pc please take a look. Commit message below. Note: I removed the part where I set `async.data` to `nullptr` because I don't think it's necessary.

---

When RTCPeerConnection's or RTCDataChannel's event loop is about to close, we
call `Unref` on either object, which signals that they can be deleted. However,
these objects hold onto the `uv_async_t` that we use for sending messages to the
event loop. If the objects are deleted before the event loop closes, we get an
error like the following:

```
node: ../deps/uv/src/unix/core.c:250: uv__finish_close: Assertion `handle->flags
& UV_CLOSING' failed.

Thread 1 "node" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) backtrace
%s%sAssertion `%s' failed.\n%n", assertion=assertion@entry=0x1761717
"handle->flags & UV_CLOSING", file=file@entry=0x17616f5
"../deps/uv/src/unix/core.c", line=line@entry=250,
function=function@entry=0x17619b0 "uv__finish_close") at assert.c:92
& UV_CLOSING", file=0x17616f5 "../deps/uv/src/unix/core.c", line=250,
function=0x17619b0 "uv__finish_close") at assert.c:101
char const* const*) ()
argv=0x7fffffffe458, init=<optimized out>, fini=<optimized out>,
rtld_fini=<optimized out>, stack_end=0x7fffffffe448) at ../csu/libc-start.c:310
```

The fix, then, is to call `Unref` in the `uv_close_cb` instead of before calling
`uv_close`.

Fixes #376

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,12 @@ New Features
 
   `portRange` cannot be updated by `setConfiguration`.
 
+Bug Fixes
+---------
+
+- Fixed a failed assertion when closing RTCPeerConnection's or RTCDataChannel's
+  event loop (#376).
+
 0.0.67
 ======
 
--- a/src/datachannel.cc
+++ b/src/datachannel.cc
@@ -220,12 +220,15 @@ void DataChannel::Run(uv_async_t* handle, int status) {
   }
 
   if (do_shutdown) {
-    self->async.data = nullptr;
-    self->Unref();
-    uv_close(reinterpret_cast<uv_handle_t*>(&self->async), nullptr);
+    uv_close(reinterpret_cast<uv_handle_t*>(&self->async), reinterpret_cast<uv_close_cb>(DataChannel::Shutdown));
   }
 
   TRACE_END;
+}
+
+void DataChannel::Shutdown(uv_async_t* handle) {
+  auto self = static_cast<DataChannel*>(handle->data);
+  self->Unref();
 }
 
 void DataChannel::OnStateChange() {
--- a/src/datachannel.h
+++ b/src/datachannel.h
@@ -97,6 +97,7 @@ class DataChannel
 
  private:
   static void Run(uv_async_t* handle, int status);
+  static void Shutdown(uv_async_t* handle);
 
   struct AsyncEvent {
     AsyncEventType type;
--- a/src/peerconnection.cc
+++ b/src/peerconnection.cc
@@ -211,12 +211,15 @@ void PeerConnection::Run(uv_async_t* handle, int status) {
   }
 
   if (do_shutdown) {
-    self->async.data = nullptr;
-    self->Unref();
-    uv_close(reinterpret_cast<uv_handle_t*>(handle), nullptr);
+    uv_close(reinterpret_cast<uv_handle_t*>(handle), reinterpret_cast<uv_close_cb>(&PeerConnection::Shutdown));
   }
 
   TRACE_END;
+}
+
+void PeerConnection::Shutdown(uv_async_t* handle) {
+  auto self = static_cast<PeerConnection*>(handle->data);
+  self->Unref();
 }
 
 void PeerConnection::OnError() {
--- a/src/peerconnection.h
+++ b/src/peerconnection.h
@@ -189,6 +189,7 @@ class PeerConnection
 
  private:
   static void Run(uv_async_t* handle, int status);
+  static void Shutdown(uv_async_t* handle);
 
   struct AsyncEvent {
     AsyncEventType type;


nazar-pc:
This doesn't work for me at all, looks like `oniceconnectionstatechange` is not fired ever.

markandrus:
Hmm, then I am suspicious of the change [here](https://github.com/js-platform/node-webrtc/pull/371) (not yet released).

nazar-pc:
Stops working on 911c9a8057f8dba88ba21fd8aead4c6aaf834b84 and works on previous commit.

nazar-pc:
Cherry-picked this PR on top of last working commit 126178e31ecc9db62ab3a1d4ecd4b0f30cbecbe5 and run my tests 10 times.
Looks great, no crashes, no errors.

markandrus:
@nazar-pc one thing that changes in 911c9a8 (and probably worth calling out in the CHANGELOG.md) is that we drop support for the (now) non-standard `url` property in RTCIceServers. The property is `urls` now. For example,

```js
new RTCPeerConnection({
  iceServers: [{ urls: 'stuns:stun.example.org' }]
});
```

Is there any chance you or one of your dependencies is using `url` instead of `urls`?

nazar-pc:
No, in tests I'm running there are no ICE servers at all, just empty array `[]` since it runs locally.

markandrus:
OK, that’s very odd. I’ll try to reproduce. Btw, will the detox-core tests repro this?

nazar-pc:
Yes, I run them unchanged, just swap `node_modules/wrtc` with symlink to cloned `node-webrtc` repository with compiled module.

nazar-pc:
Much simpler demo than `@detox/core`:
```javascript
var Peer = require('simple-peer');
var wrtc = require('.');

var peer1 = new Peer({initiator : true, wrtc, trickle: false});
var peer2 = new Peer({wrtc, trickle: false});

peer1.on('signal', function (data) {
	peer2.signal(data);
});

peer2.on('signal', function (data) {
	peer1.signal(data);
});

peer1.on('connect', function () {
	peer1.send('hey peer2, how is it going?');
});

peer2.on('data', function (data) {
	console.log('got a message from peer1: ' + data);
});
```

`trickle: false` causes issues in this case.

markandrus:
@nazar-pc OK, I'm pretty sure I know the problem. I broke simple-peer's feature detection for node-webrtc when I dropped the RTCIceConnectionStates property from RTCPeerConnection. See [here](https://github.com/feross/simple-peer/commit/2829ecef97f1b395481f497c789b221f696ecf61). I'll add it back. Not sure what a better long term solution will be. Maybe just working towards spec-compliance will be sufficient and we can remove the hacks from simple-peer. I'll open a PR for this.

> Cherry-picked this PR on top of last working commit 126178e and run my tests 10 times.
> Looks great, no crashes, no errors.

Thanks a lot for testing! Based on this, I'll merge.

nazar-pc:
Thanks for quick fix! This project is crucial for Detox existence and I greatly appreciate your work on it.

Spec compliance is always a good idea, in fact Chromium is spec compliant for most if not all WebRTC features, but there is no recent libwebrtc build that uses modern version of it yet.

nazar-pc:
If there are no known issues in develop, would be nice to see a new release with this fix soon.

PR #378, opened 2018-03-26
Fix simple-peer

markandrus:
@nazar-pc I've added back the property simple-peer uses to check for node-webrtc (although now it's an empty array). I've also enabled an existing test, `test/multiconnect.js`, which, along with `test/custom-settings.js`, uses simple-peer (and thus should catch any regressions here).


nazar-pc:
I can confirm this works for `@detox/core` too 👍

Issue #362, opened 2018-02-18
WebRTC Issue 7585: PhysicalSocketServer stops working when more than 1024 file descriptors are created


nazar-pc:
Here is a rough test case that can quite unreliably, but still reproduce the issue:
<details>

```javascript
"use strict";
const TAPE = require("tape");
const DHT = require("webtorrent-dht");

const SERVER_OPTIONS = {
	nodes: [],
	simple_peer_opts: {
		config: {
			iceServers: []
		}
	},
	isServer: true
};
const CLIENT_OPTIONS = {
	nodes: [{
		host: "127.0.0.1",
		port: 16882
	}],
	simple_peer_opts: {
		config: {
			iceServers: []
		}
	},
	isServer: false
};
//This array is used to delete all created dhts so the test exits properly.
global.running = [];
global.ready = [];


TAPE("bootstrapping test", (t) => {
	const NODES = 30;

	let p1 = new Promise((resolve) => {
		let bsDht = new DHT(SERVER_OPTIONS);
		running.push(bsDht);
		bsDht.listen(16882, "127.0.0.1", () => {});
		bsDht.once("listening", () => {
			t.pass("BS node running");
			resolve();
		});
	});

	let tmpDht = null;
	let promises = [];

	p1.then(() => {
		for(let i = 0; i < NODES; i++){
			let tmpPromise = new Promise((resolve) => {
				tmpDht = new DHT(CLIENT_OPTIONS);
				running.push(tmpDht);
				t.comment("Started Node " + i);
				ready[i] = false;
				tmpDht.once("ready", () => {
					ready[i] = true;
					t.pass("DHT ready event received " + i);
					resolve();
				});
			});
			promises.push(tmpPromise);
		}

		//Only continue if all dhts started up properly
		return Promise.all(promises);


	})
		.then(function () {
			console.log('done');
			t.end();
		})

});

TAPE.onFinish(() => {
	running.forEach((entry) => {
		entry.destroy();
	});
});
```

</details>

```
(gdb) backtrace
#0  __GI___pthread_mutex_unlock (mutex=0x7fffffffffff) at pthread_mutex_unlock.c:347
#1  0x00007fffe7675316 in rtc::PhysicalSocketServer::Wait(int, bool) () from /web/github/node-webrtc/build/Release/wrtc.node
[...]

PR #384, opened 2018-04-02
Use epoll-enabled PhysicalSocketServer on Linux

markandrus:
There may be important patches we're missing. I just copied the epoll-enabled PhysicalSocketServer from the WebRTC Issue. Needs testing.

~~**Edit:** getting crashes on Linux.~~ ~~Nevermind, I didn't set `CXX`/`CC` correctly. Seems to work?~~

**Edit 2:** Still need to define `WEBRTC_LINUX`.

---

PTAL @nazar-pc

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,11 @@ nodejs_init()
 file(GLOB_RECURSE MODULE_SRC src/*.cc src/*.h)
 add_nodejs_module(${MODULE} ${MODULE_SRC})
 
+file(GLOB_RECURSE MODULE_SRC_WITHOUT_WEBRTC src/*.cc src/*.h)
+list(REMOVE_ITEM MODULE_SRC_WITHOUT_WEBRTC
+  ${CMAKE_SOURCE_DIR}/src/webrtc/physicalsocketserver.h
+  ${CMAKE_SOURCE_DIR}/src/webrtc/physicalsocketserver.cc)
+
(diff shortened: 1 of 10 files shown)


nazar-pc:
Can't compile for some reason on Ubuntu 14.04. Upgraded Cmake from 3.2 to 3.11 (`FILTER` was not supported), but still getting error for some reason:
```
...
-- Downloading: https://nodejs.org/download/release/v9.10.1/SHASUMS256.txt
CMake Error at NodeJS.cmake:293 (message):
  Unable to extract header archive checksum
Call Stack (most recent call first):
  CMakeLists.txt:12 (nodejs_init)
...
```

UPD: Error is caused by bad libcurl, trying to resolve it

markandrus:
Sorry, @nazar-pc -- I'll have a fix for that shortly.

markandrus:
OK, I've pushed a change, and now

* The script I shared in #362 can scale up to 3000 RTCPeerConnections (much higher and I hit the thread limit on my Linux machine).
* The script I shared in #340 can scale up to 1500 RTCPeerConnection pairs (although it gets quite slow!).

nazar-pc:
Doesn't crash anymore 👍

markandrus:
M61 has the `epoll` code merged. So, once we can update libwebrtc from M60 to something newer, we can drop these extra files.

markandrus:
@nazar-pc is this OK to release from your perspective?

nazar-pc:
Absolutely!

Issue #319, opened 2017-05-25
Memory leak


NithyeshS:
I am using a fork of this repo [https://github.com/ssaroha/node-webrtc](https://github.com/ssaroha/node-webrtc) built with OpenSSL for ARM devices. I have been using it for the past few weeks and I noticed that the application memory increases as time progresses. Due to this the application fails to process(as it slows down) ICE connection checks on time and loses connectivity. I am not sure if this is an issue any body has encountered before or if it due to the building WebRTC with OpenSSL or if it an issue with the WebRTC version used. Any help regarding this is greatly appreciated.

markandrus:
The biggest leak on RTCDataChannel should be fixed now in v0.1.2.

Issue #304, opened 2017-01-30
memory leaks when transmitting data in binary format


webarchymeta:
It was found that when sending/receiving data in binary format (namely the Uint8Array format, see #303), some of the memory allocated is never released under Linux and Mac OS X, resulting in a continuously growing memory size used by the application. For Windows, the memory used can be kept at a constant, but relatively large value.

However, when encoding the data using base64 encoding, the memory used by the same application are kept at a quite smaller stable value ...

SourceBoy:
I can confirm this as well. When transferring a file, both sender & recipient's resident memory sizes increase as the file transfers, and they're not released after transfer is complete.

Given a file that's 271.8MB, when the transfer is complete, the resident memory on both ends are roughly around the size of the file (271.8MB):
```
base64 (ok):
        sender
                start { rss: 55734272, heapTotal: 32542720, heapUsed: 18060600 }
                end { rss: 71114752, heapTotal: 26251264, heapUsed: 15908624 }
        recipient
                start { rss: 56528896, heapTotal: 32542720, heapUsed: 17995136 }
                end { rss: 53059584, heapTotal: 23105536, heapUsed: 16675608 }

Uint8Array (memory issue):
        sender
                start mem { rss: 55701504, heapTotal: 32542720, heapUsed: 18082848 }
                end mem { rss: 364601344, heapTotal: 57716736, heapUsed: 29253696 }
        recipient
                start mem { rss: 56270848, heapTotal: 32542720, heapUsed: 18007416 }
                end mem { rss: 321478656, heapTotal: 22056960, heapUsed: 18232512 }
```

I've also ruled out the possibilities that this could've been caused by `Uint8Array` or Node's `stream` with separate tests using the same file with `Uint8Array` but without `RTCDataChannel.send()`:
```
Uint8Array (ok)
	sender
		start { rss: 27607040, heapTotal: 10522624, heapUsed: 4499552 }
		end { rss: 69107712, heapTotal: 49680384, heapUsed: 26788840 }
	recipient
		start { rss: 27291648, heapTotal: 10522624, heapUsed: 4385976 }
		end { rss: 40894464, heapTotal: 14716928, heapUsed: 5166192 }
```

I've previously filed a similar issue in [#205](https://github.com/js-platform/node-webrtc/issues/205#issuecomment-274725722)

vikeshkhanna:
+1

[...]

Issue #205, opened 2015-05-16
Memory leak receiving binary datachannel messages


rosskukulinski:
So I've been doing some digging into a memory leak in our application that uses node-webrtc.

We've narrowed the leak down to only when we're receiving binary datachannel messages on node 0.10.36.  When we use websockets instead of datachannels, there is no leak.

I'm not super familiar with using NAN, but I'm wondering if we're not properly cleaning up the incoming message?  https://github.com/js-platform/node-webrtc/blob/develop/src/datachannel.cc#L184

modeswitch:
Thanks for narrowing this down. I'll have a look at the code.

mafrost:
Any news on this? I'm also seeing a possible leak that we've traced to wrtc and this might be it.

SourceBoy:
I also have reasons to believe there may be memory leaks after sending/receiving data via `RTCDataChannel.send()`.

I'm using the latest 0.0.61 & linux-64 wrtc.node binary in Node 6.9.4.

Whenever I transfer a file via `send()`, the resident memory size difference between prior to transfer & after transfer is always about the size of the file, reported both by `process.memoryUsage()` and the OS itself. Closing the data channel and/or the peer connection don't seem to free the memory either, but exiting Node does.

At first I thought it could be Node's stream, but I wrote a barebone test that pipes a `fs.createReadStream()` to a `net.Socket` that sends the file to a `net.Server` with the same file, and the memory usage was fine.

markandrus:
The biggest leak on RTCDataChannel should be fixed now in v0.1.2.

PR #387, opened 2018-04-03
Refactor Events, Take 2

markandrus:
Here is another attempt at #364. IMO, it's easier to evaluate now, since other bugs have been fixed.

This should address the various memory leak issues, (#319, #304, and #205). The core issue is that we aren't deleting messages posted to the PeerConnection and DataChannel event loops. So, for example, if we receive a MessageEvent on a DataChannel, we never delete it, which means we leak the received buffer. Running @nazar-pc's script from #304 with these changes results in constant memory usage.

In order to solve this, I also took the opportunity to refactor our events into nicer, C++-style classes for processing via new classes, EventQueue and EventLoop. Using, e.g., `std::unique_ptr`s ensures we don't have to worry about manually `delete`-ing. Events are handled via double dispatch—basically, listing two from [Message Handling Without Dependencies](http://www.drdobbs.com/cpp/message-handling-without-dependencies/184429055). Per that article, there might be some opportunities to clean up boilerplate with this technique in the future.

- [x] Update DataChannel to use EventLoop
- [x] Update PeerConnection to use EventLoop

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
   ],
   "devDependencies": {
     "aws-sdk": "^2.2.10",
+    "cross-env": "^5.1.4",
     "eslint": "^4.14.0",
     "express": "^4.16.2",
     "express-browserify": "^1.0.2",
@@ -70,7 +71,7 @@
     "lint": "eslint lib/*.js lib/**/*.js test/*.js test/**/*.js examples/*.js examples/**/*.js karma.conf.js scripts/*.js",
     "pretest": "npm run wpt:init",
     "test": "node test/all.js",
-    "test:bridge": "node scripts/karma.js",
+    "test:bridge": "cross-env BROWSER=chrome node scripts/karma.js && cross-env BROWSER=firefox node scripts/karma.js",
     "wpt:init": "git submodule update --init --recursive",
     "wpt:reset": "rimraf ./test/web-platform-tests/tests && yarn wpt:init",
     "wpt:test": "mocha test/web-platform-tests/run-wpts.js",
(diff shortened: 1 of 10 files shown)


markandrus:
Chrome 63 and Karma on Appveyor has been really flaky recently. AFAICT, it's not a node-webrtc issue.

nazar-pc:
Is there anything else left to be done here? Installed some software on server and memory leaks are no fun.

markandrus:
@nazar-pc Well, it's super weird that the PeerConnection refactor is stalling the Appveyor tests. I wanted to get my Parallels license renewed so I could test it out on my laptop (rather than debug on Appveyor); if I can't get that setup later today, I might actually separate out the commits and land the first (there would still be leaks on the PeerConnection, though).

nazar-pc:
Any progress here, maybe I can help somehow?

markandrus:
@nazar-pc if you have a Windows machine and can test the second PR there, that'd help; otherwise, I'm still waiting on a Parallels license, so I've split off the first commit here: https://github.com/js-platform/node-webrtc/pull/390

nazar-pc:
I have VM with Windows 10. I don't see a list of dependencies, building process and generally I've not used Windows for anything serious in a very long time now. Could you give me short instructions?

markandrus:
For Windows, I think it should only require Microsoft Visual Studio Community Edition 2015 and CMake (although the Appveyor integration may document any other requirements, too). When I get my Windows VM setup again I'll see about documenting it.

markandrus:
The issue was actually with the bridge test, not the code changes.

PR #389, opened 2018-04-13
Updated README to use branch-head 60

williamberman:
Ran through these instructions and they worked fine. There's some hacky stuff I needed to do manually in order to cross compile. I'll look into updating https://github.com/aisouard/libwebrtc to see if it can be automated


chris-asl:
Hey @williamberman, are any of the hacky stuff mentioned in OP worth pointing out or even putting it in the README?

(Btw: You've committed a ["<<<HEAD"](https://github.com/js-platform/node-webrtc/pull/389/files#diff-04c6e90faac2675aa89e2176d2eec7d8R45) mark.)

markandrus:
#445 moves away from downloading/using aisouard/libwebrtc binaries/CMake files and instead builds WebRTC from source. This should allow passing `gn gen` arguments/cross-compiling much more readily. We could perhaps add first-class support.

Anyway, this section of the README is updated in #445, so I will close this PR.

Issue #396, opened 2018-04-22
Add support for MediaStream


markandrus:
This issue only tracks the addition of MediaStream support insofar as MediaStream is a "smart collection" of MediaStreamTracks. That is, this issue is complete once we can do the following, given some existing MediaStreamTrack, `track`,

```js
const stream = new MediaStream()
stream.addTrack(track)
```

plus any other checklist items below. `getUserMedia` is not in scope here, nor is creation of local MediaStreamTracks.

- [x] Add support for MediaStream — #398 
  - [x] constructor
    - [x] `new MediaStream()` — #398 
    - [x] `new MediaStream(stream)` — #398
    - [x] `new MediaStream(tracks)` — #398
  - [x] `id` — #398
  - [x] `getTracks` — #398
  - [x] `getAudioTracks` — #398
  - [x] `getVideoTracks` — #398
  - [x] `getTrackById` — #398
  - [x] `addTrack` — #398
  - [x] `removeTrack` — #398
  - [x] `clone`
  - [x] `active` — #398
  - [ ] `onaddtrack`
  - [ ] `onremovetrack`
- [x] Add support for `streams` in RTCTrackEvent — #398
- [x] Add support for `streams` in PeerConnection's `addTrack` — #398

avaer:
@markandrus is this list still accurate? I'd like to jump in and help but it's hard to get a lay of the land.

Specifically I'm just trying to get a mic MediaStream working and looking to what we need to finish for that.

There's also https://github.com/js-platform/node-webrtc/issues/220#issuecomment-383390412.

markandrus:
Hi, @modulesio! Yes, this last is accurate and the checked-off features have been released.

> Specifically I'm just trying to get a mic MediaStream working and looking to what we need to finish for that.

You want to acquire a MediaStream using node-webrtc that contains an audio MediaStreamTrack capturing from a microphone on the host machine? If so, we should plan out work to support `getUserMedia`. Some thoughts:

1. In order to access the microphone, we'll need to be able to create a PeerConnectionFactory that uses the default audio device for the platform node-webrtc is running on. See [here](https://github.com/js-platform/node-webrtc/blob/v0.1.6/src/peerconnectionfactory.cc#L64).
2. By default, we need to continue preferring the fake audio device.
3. Probably we should either:

    1. Parameterize the RTCPeerConnection constructor (and an eventual `getUserMedia` method) to accept the PeerConnectionFactory, or
[...]

PR #397, opened 2018-04-22
Review: thedracle/node-webrtc#mediastream-rebase-squash

markandrus:
@thedracle I'm opening this PR so that it will be easier to review which of your changes we can land. I think it may make sense to open a few separate PRs with subsets of these changes.

### DIFF
--- a/lib/mediadevices.js
+++ b/lib/mediadevices.js
@@ -1,0 +1,74 @@
+var path = require('path');
+var binary = require('node-pre-gyp');
+var path = require('path');
+//var binding_opts = require(path.resolve(path.join(__dirname, '../build/wrtc.json')));
+var binding_path = binary.find(path.resolve(path.join(__dirname,'../package.json')));
+var _webrtc = require(binding_path);
+
+var EventTarget = require('./eventtarget');
+
+function MediaDevices() {
+  'use strict';
+  var that = this
+    , md = new _webrtc.MediaDevices()
+    , localType = null
+    , remoteType = null
+    , queue = []
+    , pending = null;
+
+  EventTarget.call(this);
+
+  function checkClosed() {
+//    if(this._closed) {
+//      throw new Error('Peer is closed');
+//    }
+  }
+
+  function executeNext() {
+    if(queue.length > 0) {
+      var obj = queue.shift();
+
+      md[obj.func].apply(md, obj.args);
+
+      if(obj.wait)
+      {
+        pending = obj;
+      } else {
+        executeNext();
+      }
+    } else {
+      pending = null;
+    }
+  }
+
+  function queueOrRun(obj) {
+    console.log('------------------- queued: ' + obj.func + ', pending: ' + pending);
+    checkClosed();
+
+    if(null === pending) {
+      md[obj.func].apply(md, obj.args);
+
+      if(obj.wait) {
+        pending = obj;
+      }
+    } else {
+      queue.push(obj);
+    }
+  }
+
+  function runImmediately(obj) {
+    checkClosed();
+
+    return md[obj.func].apply(md, obj.args);
+  }
+
+  this.getUserMedia = function getUserMedia(successCallback, errorCallback){
+    var stream = md.getUserMedia();
+    if (stream === null) {
+      return errorCallback(new Error("Couldn't create MediaStream"));
+    }
+    successCallback(stream);
+  };
+}
+
+module.exports = MediaDevices;
--- a/lib/mediadevices.js
+++ b/lib/mediadevices.js
@@ -6,6 +6,7 @@ var binding_path = binary.find(path.resolve(path.join(__dirname,'../package.json
 var _webrtc = require(binding_path);
 
 var EventTarget = require('./eventtarget');
+var RTCMediaStream = require('./mediastream.js');
 
 function MediaDevices() {
   'use strict';
@@ -67,7 +68,7 @@ function MediaDevices() {
     if (stream === null) {
       return errorCallback(new Error("Couldn't create MediaStream"));
     }
-    successCallback(stream);
+    successCallback(new RTCMediaStream(stream));
   };
 }
 
--- a/lib/mediastream.js
+++ b/lib/mediastream.js
@@ -1,0 +1,61 @@
+var EventTarget = require('./eventtarget');
+
+var RTCMediaStreamMessageEvent = require('./mediastreammessageevent');
+
+function RTCMediaStream(internalMS) {
+  'use strict';
+  var that = this;
+
+  EventTarget.call(this);
+
+  that._internalMS = internalMS;
+  internalMS.onerror = function onerror() {
+    that.dispatchEvent({type: 'error'});
+  };
+
+  internalMS.onmessage = function onmessage(data) {
+    var evt = new RTCMediaStreamMessageEvent(data);
+    that.dispatchEvent(evt);
+    that._onmessage && that._onmessage(evt);
+  };
+
+  that.getVideoTracks = function() {
+    return internalMS.getVideoTracks();
+  }
+
+  Object.defineProperties(this, {
+    'id': {
+        get: function getID() {
+            return internalMS.id;
+        }
+    },
+    'active': {
+        get: function() {
+          return internalMS.active;
+        }
+    },
+    'ended': {
+        get: function() {
+          return internalMS.readyState === 'closed';
+        }
+    },
+    'onmessage': {
+        set: function(onMessage) {
+            that._onmessage = onMessage;
+        }
+    }
+  });
+
+  this.close = function close() {
+    internalMS.close();
+  };
+}
+
+RTCMediaStream.prototype.RTCMediaStates = [
+  'connecting',
+  'open',
+  'closing',
+  'closed'
+];
+
+module.exports = RTCMediaStream;
--- a/lib/mediastream.js
+++ b/lib/mediastream.js
@@ -19,10 +19,15 @@ function RTCMediaStream(internalMS) {
     that._onmessage && that._onmessage(evt);
   };
 
+  that.getAudioTracks = function() {
+    return internalMS.getAudioTracks();
+  }
   that.getVideoTracks = function() {
     return internalMS.getVideoTracks();
   }
-
+  that.getTracks = function() {
+    return internalMS.getVideoTracks().concat(internalMS.getAudioTracks());
+  }
   Object.defineProperties(this, {
     'id': {
         get: function getID() {
--- a/lib/mediastreamevent.js
+++ b/lib/mediastreamevent.js
@@ -1,0 +1,15 @@
+function RTCMediaStreamEvent(type, eventInitDict) {
+  'use strict';
+  Object.defineProperties(this, {
+    'type': {
+      value: type,
+      enumerable: true
+    },
+    'stream': {
+      value: eventInitDict.stream,
+      enumerable: true
+    }
+  });
+}
+
+module.exports = RTCMediaStreamEvent;
--- a/lib/peerconnection.js
+++ b/lib/peerconnection.js
@@ -6,10 +6,13 @@ var EventTarget = require('./eventtarget');
 
 var RTCDataChannel = require('./datachannel');
 var RTCDataChannelEvent = require('./datachannelevent');
+var RTCMediaStreamEvent       = require('./mediastreamevent');
 var RTCIceCandidate = require('./icecandidate');
 var RTCPeerConnectionIceEvent = require('./rtcpeerconnectioniceevent');
 var RTCSessionDescription = require('./sessiondescription');
 var RTCStatsResponse = require('./rtcstatsresponse');
+var RTCMediaStream            = require('./mediastream');
+var RTCVideoFrameEvent        = require('./rtcvideoframeevent');
 
 function RTCPeerConnection(configuration, constraints) {
   var self = this;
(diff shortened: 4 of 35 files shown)


lib/mediastream.js
```diff
@@ -0,0 +1,66 @@
...
+}
+
+RTCMediaStream.prototype.RTCMediaStates = [
+  'connecting',
+  'open',
+  'closing',
+  'closed'
+];
+
+module.exports = RTCMediaStream;
```
markandrus, line 66: This looks roughly like what we will need for MediaStream. In #393, I've added MediaStreamTrack and tried to avoid the JavaScript wrapper (everything's implemented in the C++) layer; however, I haven't added—nor do I have a good pattern for—EventTarget support (which we need for both MediaStream and MediaStreamTrack). Until we can handle EventTarget support in C++, I think we can keep the JavaScript wrappers.

lib/mediadevices.js
```diff
@@ -0,0 +1,75 @@
...
+  this.getUserMedia = function getUserMedia(successCallback, errorCallback){
+    var stream = md.getUserMedia();
+    if (stream === null) {
+      return errorCallback(new Error("Couldn't create MediaStream"));
+    }
+    successCallback(new RTCMediaStream(stream));
+  };
+}
+
+module.exports = MediaDevices;
```
markandrus, line 75: I think it should be possible to remove the JavaScript wrapper here? Also, I don't think we will need the `executeNext`/`queueOrRun` infrastructure going forward. It's been removed from `lib/peerconnection.js`, for example.

lib/mediastreamevent.js
```diff
@@ -0,0 +1,15 @@
...
+      enumerable: true
+    },
+    'stream': {
+      value: eventInitDict.stream,
+      enumerable: true
+    }
+  });
+}
+
+module.exports = RTCMediaStreamEvent;
```
markandrus, line 15: We may just want to use RTCTrackEvent (implemented in #393), since `onaddstream` is not present in the recent WebRTC draft spec. We could then consider shimming legacy APIs in the JavaScript layer.

lib/peerconnection.js
```diff
@@ -81,39 +85,145 @@ function RTCPeerConnection(configuration, constraints) {
...
+    self.dispatchEvent(videoFrameEvent);
+    self._onvideoframe && self._onvideoframe(videoFrameEvent);
+  };
+
+  pc.onencodedvideoframe = function(id, width, height, encodedFrame, frameType, timestamp) {
+      var videoFrameEvent = new RTCVideoFrameEvent('videoframe',
+          {id: id, width: width, height: height, encodedFrame: encodedFrame,
+              frameType: frameType, timestamp: timestamp});
+      self.dispatchEvent(videoFrameEvent);
+      self._onencodedvideoframe && self._onencodedvideoframe(videoFrameEvent);
```
markandrus, line 159: I think an issue with the `onaudioframe`/`onvidoeframe`/`onencodedvideoframe` will be that it doesn't make it easy to distinguish frames for different tracks of the same kind? Unless `id` will match the underlying MediaStreamTrack's `id`? I'm not sure how this works yet. **Edit:** I see below now it does.

IMO, it would be better if these events were raised on the MediaStreamTracks themselves, rather than the RTCPeerConnection. IIUC, `onencodedvideoframe` should make sense on a local video MediaStreamTrack even if its not yet added to an RTCPeerConnection.

lib/peerconnection.js
```diff
@@ -340,7 +450,38 @@ function RTCPeerConnection(configuration, constraints) {
...
+              typeof(arguments[1]) === 'function' &&
+              typeof(arguments[2]) === 'function') {
+              // Legacy method.
+              call(stream, arguments[1], arguments[2]);
+          }
+          else {
+              throw error;
+          }
+      }
+  }
```
markandrus, line 482: Let us use `addTrack`, once it is supported, and shim `addStream` in terms of that (or delegating to libwebrtc, if it does the shimming itself). But, in either situation, this method should be `void` AFAICT.

lib/rtcaudioframeevent.js
```diff
@@ -0,0 +1,35 @@
...
+            enumerable: false
+        },
+        'buffer': {
+            value: eventInitDict.buffer,
+            enumerable: true
+        }
+    });
+}
+
+module.exports = RTCAudioFrameEvent;
```
markandrus, line 35: This looks useful 👍we may want to use camelCase for the attributes, though.

lib/rtcvideoframeevent.js
```diff
@@ -0,0 +1,45 @@
...
+            enumerable: true
+        },
+        'vPlane': {
+            value: eventInitDict.vPlane,
+            enumerable: true
+        }
+    });
+}
+
+module.exports = RTCVideoFrameEvent;
```
markandrus, line 45: This also looks useful 👍same comment as above.

src/decoderproxy.h
```diff
@@ -0,0 +1,43 @@
...
+#define NODEVP8_DECODER_PROXY_H
+
+#include "webrtc/modules/video_coding/codecs/vp8/include/vp8.h"  // nogncheck
+#include "webrtc/modules/video_coding/utility/ivf_file_writer.h"
+#include <memory>
+#include <string>
+#include <functional>
+#include <map>
+
+class DecoderProxy : public webrtc::VideoDecoder {
```
markandrus, line 11: Why is this necessary? I may have to read up...

src/mediadevices.cc
```diff
@@ -0,0 +1,99 @@
...
+using v8::String;
+using v8::Uint32;
+using v8::Value;
+
+Nan::Persistent<Function> MediaDevices::constructor;
+
+
+// MediaDevices constructor/destructor
+MediaDevices::MediaDevices() {
+  _peer_connection_factory = PeerConnectionFactory::GetOrCreateDefault();
```
markandrus, line 30: PeerConnectionFactory should allow choosing the AudioLayer it uses to `Create` AudioDeviceModule. Once we do this, it may make sense to pass along the PeerConnectionFactory (specifying a particular AudioLayer) to classes like PeerConnection and MediaDevices.

We probably also want a static `getUserMedia` that lazily constructs MediaDevices, not unlike PeerConnection, if possible? Not sure if that'd complicate the code, but it'd mean

```js
const { getUserMedia } = require('wrtc')

(async () => {
  const stream = getUserMedia({ audio: true, video: true })
  // PeerConnectionFactory constructed
  stream.getTracks().forEach(track => track.stop())
})().then(() => {
  // PeerConnectionFactory destroyed
})
```

just as the following works today:

```js
const { RTCPeerConnection } = require('wrtc')

(async () => {
  const pc = new RTCPeerConnection()
  // PeerConnectionFactory constructed
  pc.close()
})().then(() => {
  // PeerConnectionFactory destroyed
})
```

src/mediadevices.cc
```diff
@@ -0,0 +1,99 @@
...
+}
+
+NAN_METHOD(MediaDevices::GetUserMedia) {
+  TRACE_CALL;
+
+  MediaDevices* self = Nan::ObjectWrap::Unwrap<MediaDevices>(info.This());
+
+  const char kAudioLabel[] = "audio_label";
+//  const char kVideoLabel[] = "video_label";
+  const char kStreamLabel[] = "stream_label";
```
markandrus, line 63: We'll want to generate UUIDs, I think.

src/mediadevices.cc
```diff
@@ -0,0 +1,99 @@
...
+      self->_peer_connection_factory->factory()->CreateLocalMediaStream(kStreamLabel);
+
+  stream->AddTrack(audio_track);
+//  stream->AddTrack(video_track);
+
+  Local<Value> cargv[1];
+  cargv[0] = Nan::New<External>(static_cast<void*>(stream));
+  Local<Value> obj = Nan::New(MediaStream::constructor)->NewInstance(1, cargv);
+
+  info.GetReturnValue().Set(obj);
```
markandrus, line 85: We should construct a `v8::Promise::Resolver`, call `Resolve` with the MediaStream, and then return the result of calling `GetPromise` on the `v8::Promise::Resolver`.

src/mediadevices.cc
```diff
@@ -0,0 +1,99 @@
...
+
+  TRACE_END;
+}
+
+void MediaDevices::Init(Handle<Object> exports) {
+  Local<FunctionTemplate> tpl = Nan::New<FunctionTemplate>(New);
+  tpl->SetClassName(Nan::New("MediaDevices").ToLocalChecked());
+  tpl->InstanceTemplate()->SetInternalFieldCount(1);
+
+  Nan::SetPrototypeMethod(tpl, "getUserMedia", GetUserMedia);
```
markandrus, line 95: Seems like we'll want this (so that you could have two MediaDevices using different AudioLayers), although we may want a static `getUserMedia` that uses the default PeerConnectionFactory (see comment above).

src/mediastream.cc
```diff
@@ -0,0 +1,270 @@
...
+using v8::Uint32;
+using v8::Value;
+
+Nan::Persistent<Function> MediaStream::constructor;
+
+MediaStream::MediaStream(webrtc::MediaStreamInterface* msi)
+  : loop(uv_default_loop()),
+    _internalMediaStream(msi) {
+  uv_mutex_init(&lock);
+  uv_async_init(loop, &async, reinterpret_cast<uv_async_cb>(Run));
```
markandrus, line 33: We'll want to convert to EventLoop rather than managing libuv directly.

src/mediastream.cc
```diff
@@ -0,0 +1,270 @@
...
+
+    }
+    if(MediaStream::ADDTRACK & evt.type) 
+    {
+
+    }
+    if(MediaStream::REMOVETRACK & evt.type) 
+    {
+
+    }
```
markandrus, line 119: One of my questions with adding MediaStream (related to #396) is this: when a MediaStreamTrackInterface is added to a MediaStreamInterface by WebRTC, how do we figure out if it needs to be wrapped in our own MediaStreamTrack wrapper or not? We don't just want to re-wrap in a new MediaStreamTrack every time. Maybe the MediaStreamTrack class should have a static map from MediaStreamTrackInterfaces to MediaStreamTracks, and we just do a "get or create" operation for a new MediaStreamTrack? Then I'm just worried about leaks, but probably these can be removed in response to some event.

src/mediastream.cc
```diff
@@ -0,0 +1,270 @@
...
+                                   == track->GetSource()->kMuted));
+
+    // ToDO: Determine if track is read only.
+    Nan::Set(jsAudioTrack, Nan::New<v8::String>("readonly").ToLocalChecked(),
+             Nan::New<v8::Boolean>(true));
+
+    Nan::Set(jsAudioTrack, Nan::New<v8::String>("remote").ToLocalChecked(),
+             Nan::New<v8::Boolean>(track->GetSource()->remote()));
+
+    Nan::Set(audioTracksJSArray, Nan::New<Number>(i++), jsAudioTrack);
```
markandrus, line 165: We should be able to update this to use the new `CONVERT_OR_THROW_AND_RETURN` macro.

src/mediastream.cc
```diff
@@ -0,0 +1,270 @@
...
+                                   == track->GetSource()->kMuted));
+
+    // ToDO: Determine if track is read only.
+    Nan::Set(jsVideoTrack, Nan::New<v8::String>("readonly").ToLocalChecked(),
+             Nan::New<v8::Boolean>(true));
+
+    Nan::Set(jsVideoTrack, Nan::New<v8::String>("remote").ToLocalChecked(),
+             Nan::New<v8::Boolean>(track->GetSource()->remote()));
+
+    Nan::Set(videoTracksJSArray, Nan::New<Number>(i++), jsVideoTrack);
```
markandrus, line 202: Same comment as above.

src/mediastreamtrack.cc
```diff
@@ -0,0 +1,216 @@
...
+}
+
+NAN_METHOD(MediaStreamTrack::stop) {
+  TRACE_CALL;
+
+  // todo: implement
+
+  TRACE_END;
+  info.GetReturnValue().Set(Nan::Undefined());
+}
```
markandrus, line 103: The implementation in #393 lacks support for this. We should add it.

src/mediastreamtrack.cc
```diff
@@ -0,0 +1,216 @@
...
+
+NAN_GETTER(MediaStreamTrack::GetLabel) {
+  TRACE_CALL;
+
+  MediaStreamTrack* self = Nan::ObjectWrap::Unwrap<MediaStreamTrack>( info.Holder() );
+  std::string label = self->_internalMediaStreamTrack->id();
+
+  TRACE_END;
+  info.GetReturnValue().Set(Nan::New(label.c_str()).ToLocalChecked());
+}
```
markandrus, line 123: The implementation in #393 lacks support for this. We should add it.

src/mediastreamtrack.cc
```diff
@@ -0,0 +1,216 @@
...
+
+NAN_METHOD(MediaStreamTrack::clone) {
+  TRACE_CALL;
+
+  // todo: implement
+
+  TRACE_END;
+  info.GetReturnValue().Set(Nan::Undefined());
+}
```
markandrus, line 94: The implementation in #393 lacks support for this. We should add it.

src/mediastreamtrack.cc
```diff
@@ -0,0 +1,216 @@
...
+
+NAN_GETTER(MediaStreamTrack::GetMuted) {
+  TRACE_CALL;
+
+  // todo: implement
+
+  TRACE_END;
+  info.GetReturnValue().Set(Nan::New(false));
+}
```
markandrus, line 153: The implementation in #393 lacks support for this. We should add it.

src/peerconnection.cc
```diff
@@ -181,6 +192,155 @@ void PeerConnection::Run(uv_async_t* handle, int status) {
...
+            }
+
+            self->QueueEvent(PeerConnection::NOTIFY_ON_AUDIO_FRAME, audioFrameEvent);
+
+            audioSampleBuffer->assign(audioSampleBuffer->begin() + AUDIO_SAMPLE_DELIVERY_SIZE, audioSampleBuffer->end());
+          }
+        }, label);
+        audioTrack->GetSource()->AddSink(sink);
+      }
+    }
```
markandrus, line 339: We should convert these to `Handle*` methods (newer style introduced in #387).

src/peerconnection.cc
```diff
@@ -193,6 +353,99 @@ void PeerConnection::Run(uv_async_t* handle, int status) {
...
+  auto videoTracks = mediaStreamInterface->GetVideoTracks();
+  string label = mediaStreamInterface->label();
+  for(auto videoTrack : videoTracks) {
+    videoTrack->AddOrUpdateSink(new rtc::RefCountedObject<VideoSink>([label, self](const webrtc::VideoFrame& frame, string inLabel) {
+      VideoFrameEvent* event = new VideoFrameEvent(label);
+      event->width = frame.width();
+      event->height = frame.height();
+      event->buffer = frame.video_frame_buffer();
+      self->QueueEvent(PeerConnection::NOTIFY_ON_FRAME, event);
+    }, label), rtc::VideoSinkWants());
```
markandrus, line 375: How should this sink be removed?

src/peerconnection.cc
```diff
@@ -193,6 +353,99 @@ void PeerConnection::Run(uv_async_t* handle, int status) {
...
+
+      event->timestamp = bigPts;
+      wrapAroundData.lastTimeStamp = event->timestamp;
+
+      event->width = frame._encodedWidth;
+      event->height = frame._encodedHeight;
+
+      event->buffer.insert(event->buffer.begin(), frame._buffer, frame._buffer + frame._length);
+      self->QueueEvent(PeerConnection::NOTIFY_ON_ENCODED_FRAME, event);
+    });
```
markandrus, line 442: Similar question as above: should this be/is this unregistered?

lib/mediastream.js
thedracle: Nice. I have a sort of mockup of MediaStreamTrack in my branch that just composes an JS object from, but it definitely seems better to avoid the wrapper.

Just a question--- does this make it so in javascript land a track would pass type checks against RTCMediaStreamTrack?

There is some nicety in being able to wrap and modify the interface to these things in Javascript rather than in C++, particularly if in the future we wanted to more easily provide support for spec changes without having to rewrite a significant portion of the C++ interface.

markandrus: > Just a question--- does this make it so in javascript land a track would pass type checks against RTCMediaStreamTrack?

Sorry, I'm not sure what you mean. Can you ask in a different way? With either approach, given some `mediaStreamTrack` and `mediaStream`, we can satisfy

```js
const { MediaStream, MediaStreamTrack } = require('wrtc')
mediaStream instanceof MediaStream
mediaStreamTrack instanceof MediaStreamTrack
```

if we want to. We can also ensure methods that should only accept proper instances of these accept only proper instances, too. There's just more wrapping and unwrapping necessary if we have JavaScript wrappers.

> There is some nicety in being able to wrap and modify the interface to these things in Javascript rather than in C++, particularly if in the future we wanted to more easily provide support for spec changes without having to rewrite a significant portion of the C++ interface.

Yes, this is true. For example, it might be nice to shim the legacy APIs in JavaScript rather than bind to the native APIs (I'm thinking `addStream`, or the callback versions of `setLocalDescription`, etc.—unless the shims are complicated enough and libwebrtc actually provides correct legacy APIs). As far as newly specced APIs, I'm not sure what's easier in JavaScript right now. But we _can_ still shim methods C++ classes' `prototype` if necessary.

markandrus:
Although different, I think #472 enables similar use cases as this PR, at least for video. We can track audio work in #478. Although I didn't land changes from this PR, it was useful to see the implementation.

PR #398, opened 2018-04-23
RTCRtpSender and MediaStream support

markandrus:
_Not ready to merge._

See checklist in #396.

One thing to figure out how to do: raising `onaddtrack` and `onremovetrack` on the MediaStream. So far, objects that raise events extend from EventLoop; but these need to be stopped via Stop. There's no clear event that should trigger Stop, at least for both the local and remote MediaStreams. For remote MediaStreams this could coincide with closing the RTCPeerConnection; but I'm not sure about local MediaStreams. Perhaps MediaStream's private constructor could accept a boolean indicating whether it is local or not, and then we let local MediaStreams stop themselves immediately? That could solve the problem, although it feels a little weird. I think the alternative would require RTCPeerConnection driving events on remote MediaStreams it "owns" itself. The annoying thing, though, is that the observer for the MediaStream would get registered on the RTCPeerConnection rather than the MediaStream, and then the RTCPeerConnection would have to reach into the MediaStream to raise the events. Perhaps the former is cleaner then.

### DIFF
--- a/src/converters/arguments.h
+++ b/src/converters/arguments.h
@@ -23,10 +23,23 @@
 
 namespace node_webrtc {
 
+struct Arguments {
+  Nan::NAN_METHOD_ARGS_TYPE info;
+  explicit Arguments(Nan::NAN_METHOD_ARGS_TYPE info): info(info) {}
+};
+
 template <typename A>
-struct Converter<Nan::NAN_METHOD_ARGS_TYPE, A> {
-  static Validation<A> Convert(Nan::NAN_METHOD_ARGS_TYPE info) {
-    return From<A>(info[0]);
+struct Converter<Arguments, A> {
+  static Validation<A> Convert(Arguments args) {
+    return From<A>(args.info[0]);
+  }
+};
+
+template <typename L, typename R>
+struct Converter<Arguments, Either<L, R>> {
+  static Validation<Either<L, R>> Convert(Arguments args) {
+    return From<L>(args).Map(&Either<L, R>::Left)
+        | (From<R>(args).Map(&Either<L, R>::Right));
   }
 };
 
@@ -36,11 +49,11 @@ static std::tuple<A, B> Make2Tuple(A a, B b) {
 }
 
 template <typename A, typename B>
-struct Converter<Nan::NAN_METHOD_ARGS_TYPE, std::tuple<A, B>> {
-  static Validation<std::tuple<A, B>> Convert(Nan::NAN_METHOD_ARGS_TYPE info) {
+struct Converter<Arguments, std::tuple<A, B>> {
+  static Validation<std::tuple<A, B>> Convert(Arguments args) {
     return curry(Make2Tuple<A, B>)
-        % From<A>(info[0])
-        * From<B>(info[1]);
+        % From<A>(args.info[0])
+        * From<B>(args.info[1]);
   }
 };
 
@@ -50,12 +63,12 @@ static std::tuple<A, B> Make3Tuple(A a, B b, C c) {
 }
 
 template <typename A, typename B, typename C>
-struct Converter<Nan::NAN_METHOD_ARGS_TYPE, std::tuple<A, B, C>> {
-  static Validation<std::tuple<A, B, C>> Convert(Nan::NAN_METHOD_ARGS_TYPE info) {
+struct Converter<Arguments, std::tuple<A, B, C>> {
+  static Validation<std::tuple<A, B, C>> Convert(Arguments args) {
     return curry(Make3Tuple<A, B, C>)
-        % From<A>(info[0])
-        * From<B>(info[1])
-        * From<C>(info[2]);
+        % From<A>(args.info[0])
+        * From<B>(args.info[1])
+        * From<C>(args.info[2]);
   }
 };
 
--- a/src/converters/webrtc.h
+++ b/src/converters/webrtc.h
@@ -25,6 +25,7 @@
 #include "src/mediastream.h"
 #include "src/mediastreamtrack.h"
 #include "src/rtcrtpreceiver.h"
+#include "src/rtcrtpsender.h"
 
 namespace node_webrtc {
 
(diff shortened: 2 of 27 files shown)


src/converters.h
```diff
@@ -63,8 +63,8 @@ struct Converter<T, T> {
 template <typename S, typename L, typename R>
 struct Converter<S, Either<L, R>> {
   static Validation<Either<L, R>> Convert(const S s) {
-    return From<L>(s).Map(&Either<L, R>::Left)
-        | (From<R>(s).Map(&Either<L, R>::Right));
+    return Converter<S, L>::Convert(s).Map(&Either<L, R>::Left)
+        | (Converter<S, R>::Convert(s).Map(&Either<L, R>::Right));
```
markandrus: I hit some issues with ambiguous template specialization. Might be able to revert to the nicer-to-read version.

src/converters/arguments.h
```diff
@@ -23,10 +23,25 @@
 
 namespace node_webrtc {
 
+struct Arguments {
+ public:
+  Nan::NAN_METHOD_ARGS_TYPE info;
+  explicit Arguments(Nan::NAN_METHOD_ARGS_TYPE info): info(info) {}
+};
```
markandrus: I hit some issues with ambiguous template specialization. Might be able to revert this change.

src/converters/webrtc.h
```diff
@@ -620,6 +623,48 @@ struct Converter<webrtc::MediaStreamTrackInterface::TrackState, std::string> {
...
+
+template <>
+struct Converter<v8::Local<v8::Value>, node_webrtc::MediaStreamTrack*> {
+  static Validation<node_webrtc::MediaStreamTrack*> Convert(v8::Local<v8::Value> value) {
+    // TODO(mroberts): This is not safe.
+    return value->IsObject() && !value->IsArray()
+        ? Validation<node_webrtc::MediaStreamTrack*>::Valid(Nan::ObjectWrap::Unwrap<node_webrtc::MediaStreamTrack>(value->ToObject()))
+        : Validation<node_webrtc::MediaStreamTrack*>::Invalid("IDK");
+  }
+};
```
markandrus: The safe way to do this, and the conversion above, would be something like this: https://stackoverflow.com/questions/8994196/how-to-check-for-correct-type-when-calling-objectwrapunwrap-in-a-nodejs-add-on

src/mediastream.cc
```diff
@@ -0,0 +1,243 @@
...
+  auto self = Nan::ObjectWrap::Unwrap<MediaStream>(info.Holder());
+  auto tracks = std::vector<MediaStreamTrack*>();
+  for (auto const& track : self->_stream->GetAudioTracks()) {
+    auto mediaStreamTrack = MediaStreamTrack::GetOrCreate(self->_factory, track);
+    tracks.push_back(mediaStreamTrack);
+  }
+  for (auto const& track : self->_stream->GetVideoTracks()) {
+    auto mediaStreamTrack = MediaStreamTrack::GetOrCreate(self->_factory, track);
+    tracks.push_back(mediaStreamTrack);
+  }
```
markandrus: This is used in a couple places. Worth factoring out into a `tracks` method. Then we can reuse in `GetTracks`, `GetActive`, `GetTrackById`, etc.

src/mediastream.cc
```diff
@@ -0,0 +1,243 @@
...
+  if (track->kind() == track->kAudioKind) {
+    stream->RemoveTrack(static_cast<webrtc::AudioTrackInterface*>(track.get()));
+  } else {
+    stream->RemoveTrack(static_cast<webrtc::VideoTrackInterface*>(track.get()));
+  }
+}
+
+NAN_METHOD(MediaStream::Clone) {
+  Nan::ThrowError("Not yet implemented; file a feature request against node-webrtc");
+}
```
markandrus: Implement this by returning the C++ equivalent of `new MediaStream(this)`.

src/mediastream.h
```diff
@@ -0,0 +1,67 @@
...
+ public:
+  MediaStream(
+      std::shared_ptr<node_webrtc::PeerConnectionFactory>&& factory,
+      rtc::scoped_refptr<webrtc::MediaStreamInterface>&& stream):
+    MediaStream(std::move(factory), std::move(stream), !_shouldReleaseFactory) {}
+
+  MediaStream(
+      std::shared_ptr<node_webrtc::PeerConnectionFactory>&& factory,
+      rtc::scoped_refptr<webrtc::MediaStreamInterface>&& stream,
+      bool shouldReleaseFactory);
```
markandrus: Would be good to comment on these. The reason you might pass a non-null `factory` _and_ set `shouldReleaseFactory` to `true` are a result of the `new MediaStream(tracks)` constructor and needing to use the default factory. Actually, maybe we could still encapsulate this logic inside the "smaller" constructor when `factory` and `stream` are `nullptr`.

aldahick:
Forgive me if this is obvious, but I'm not sure what the progress on this implementation is. Is it still being blocked by `onaddtrack` and `onremovetrack`? As many other people have mentioned in other places, this will be great once it's published!

markandrus:
Hey @aldahick,

I saw a crash for one of the Appveyor builds. I want to try to replicate in a Windows VM and fix it before merging. I’ve just been busy recently! I think I’ll have some time this week/weekend.

markandrus:
I looked into the bug. It's reproducible on Linux on macOS, too, in Node 9.8.0, AFAICT. I'm not sure if it's a Nan bug or something wrong with the way I'm using Nan/v8 and libuv APIs. I created a minimal repro example [here](https://github.com/markandrus/nan-asyncresource-bug) and opened up an issue on nodejs/nan.

markandrus:
For reference, the bug I filed is here: https://github.com/nodejs/nan/issues/772 **EDIT**: I merged #407 which gives a new pattern for handling AsyncResource + ObjectWrap and rebased this branch.

markandrus:
Answering my own questions:

> One thing to figure out how to do: raising `onaddtrack` and `onremovetrack` on the MediaStream.

`onaddtrack` and `onremovetrack` will only be raised on remote MediaStreams. the PeerConnection class, when handling the `ontrack` event, can raise the event on MediaStream.

However, I would like to punt on `onaddtrack` and `onremovetrack` for now so that I can merge this PR sooner than later. In my testing, things seem pretty good. I don't believe there are any memory leaks with the new objects. We also have destructor tests now to ensure there are no leaks.

Issue #358, opened 2018-02-09
Connections closing error


bpatrik:
Hi,

I have a node backend and normal webpage fronted running in chrome. They communicate over WebRTC. 
I experienced some error when I close a WebRTC connection on the node side. Mostly when I close to early, probably before the connection is properly established. 
I could reproduce the error when I was opening and closing the connections (by simple reloading the frontend page quickly)
Unfortunately, the only error description I have is this:
`Process finished with exit code -1073741819 (0xC0000005)`
I'm using  `v0.0.66`  of the library. 

Thank you!

markandrus:
Thanks for the report @bpatrik. Is there any code you could share so that I could reproduce? Also, it should be possible to do something like

```
$ lldb -e node
> run index.js
# crash
> bt
```

to get a backtrace (or with gdb).

bpatrik:
This is what `lldb` gives me on my code:
```
Process 15639 stopped
* thread #1: tid = 15639, 0x0000000001418cb1 node`uv__async_io(loop=0x00000000021b5760, w=0x0000000000000001, events=36163120) + 209 at async.c:109, name = 'node', stop reason = signal SIGSEGV: invalid address (fault address: 0x8)
    frame #0: 0x0000000001418cb1 node`uv__async_io(loop=0x00000000021b5760, w=0x0000000000000001, events=36163120) + 209 at async.c:109
(lldb) bt
* thread #1: tid = 15639, 0x0000000001418cb1 node`uv__async_io(loop=0x00000000021b5760, w=0x0000000000000001, events=36163120) + 209 at async.c:109, name = 'node', stop reason = signal SIGSEGV: invalid address (fault address: 0x8)
  * frame #0: 0x0000000001418cb1 node`uv__async_io(loop=0x00000000021b5760, w=0x0000000000000001, events=36163120) + 209 at async.c:109
    frame #1: 0x000000000142ac58 node`uv__io_poll(loop=0x00000000021b5760, timeout=602) + 952 at linux-core.c:400
    frame #2: 0x0000000001419676 node`uv_run(loop=0x00000000021b5760, mode=UV_RUN_DEFAULT) + 342 at core.c:368
    frame #3: 0x00000000008d49ed node`node::Start(uv_loop_s*, int, char const* const*, int, char const* const*) + 1213
    frame #4: 0x00000000008d12d3 node`node::Start(int, char**) + 339
[...]

PR #418, opened 2018-06-28
fix race during close

sudison:
It's related to issue: https://github.com/js-platform/node-webrtc/issues/358

### DIFF
--- a/lib/datachannel.js
+++ b/lib/datachannel.js
@@ -17,8 +17,8 @@ function RTCDataChannel(internalDC) {
     self.dispatchEvent(new RTCDataChannelMessageEvent(data));
   };
 
-  internalDC.onstatechange = function onstatechange() {
-    switch (this.readyState) {
+  internalDC.onstatechange = function onstatechange(data) {
+    switch (data) {
       case 'open':
         self.dispatchEvent({ type: 'open' });
         break;
--- a/src/datachannel.cc
+++ b/src/datachannel.cc
@@ -108,7 +108,14 @@ void DataChannel::HandleErrorEvent(const ErrorEvent<DataChannel>& event) {
 void DataChannel::HandleStateEvent(const DataChannelStateChangeEvent& event) {
   TRACE_CALL;
   Nan::HandleScope scope;
-  MakeCallback("onstatechange", 0, nullptr);
+  Local<Value> argv[1];
+  if (event.state == webrtc::DataChannelInterface::kClosed) {
+    argv[0] = Nan::New("closed").ToLocalChecked();
+  } else {
+    argv[0] = Nan::New("open").ToLocalChecked();
+  }
+  
+  MakeCallback("onstatechange", 1, argv);
   if (event.state == webrtc::DataChannelInterface::kClosed) {
     Stop();
   }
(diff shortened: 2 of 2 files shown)


markandrus:
Thank you for this, @sudison. Is there any chance you could add a test demonstrating the issue that fails before your change and is resolved by your change? That would ensure we do not regress here.

sudison:
It's a race condition, may not be able to write test cases to reproduce the issue easily.

markandrus:
@sudison OK, I am still trying to understand the change then. From looking at the excerpts below, the JavaScript changes should be no different than what we had before (since `this.readyState` is just a getter for `internalDC.readyState`) _unless_ it is the case that HandleStateEvent can be called with a state that does not match the underlying `_jingleDataChannel->state()`? i.e., `this.readyState` is still "open" but we got an `event` that says "closed"? Does this hypothesis make sense and is it the one you were working under?

Thinking some more, I do like this change, because my goal is that we can eventually remove the JavaScript layer. Therefore, the C++ RTCDataChannel should only define APIs in the WebRTC spec. So eventually, I'd like `internalDC.onstatechange` to be removed, and we just surface `onopen` and `onclose`, per [the spec](https://w3c.github.io/webrtc-pc/#rtcdatachannel).

**datachannel.js**

```js
-  internalDC.onstatechange = function onstatechange() {
-    switch (this.readyState) {
+  internalDC.onstatechange = function onstatechange(data) {
+    switch (data) {
       case 'open':
         self.dispatchEvent({ type: 'open' });
         break;
       case 'closed':
         self.dispatchEvent({ type: 'close' });
         break;
     }
   };
```

```js
// ...
    readyState: {
      get: function getReadyState() {
        return internalDC.readyState;
      }
    },
// ...
```

**datachannel.cc**

```diff
 void DataChannel::HandleStateEvent(const DataChannelStateChangeEvent& event) {
   TRACE_CALL;
   Nan::HandleScope scope;
-  MakeCallback("onstatechange", 0, nullptr);
+  Local<Value> argv[1];
+  if (event.state == webrtc::DataChannelInterface::kClosed) {
+    argv[0] = Nan::New("closed").ToLocalChecked();
+  } else {
+    argv[0] = Nan::New("open").ToLocalChecked();
+  }
+  
+  MakeCallback("onstatechange", 1, argv);
   if (event.state == webrtc::DataChannelInterface::kClosed) {
     Stop();
   }
   TRACE_END;
 }
```

```cpp
NAN_GETTER(DataChannel::GetReadyState) {
  TRACE_CALL;
  (void) property;

  auto self = AsyncObjectWrapWithLoop<DataChannel>::Unwrap(info.Holder());

  CONVERT_OR_THROW_AND_RETURN(self->_jingleDataChannel
      ? self->_jingleDataChannel->state()
      : webrtc::DataChannelInterface::kClosed,
      state,
      Local<Value>);

  TRACE_END;
  info.GetReturnValue().Set(state);
}
```

sudison:
Yes, we are on the same page about the race condition in GetReadyState.

markandrus:
OK, thanks @sudison 👍

markandrus:
Do you mind running `make format`? That should cause Travis to pass. Then we can merge.

PR #424, opened 2018-08-25
Update package dependencies

Dreaded-Gnu:
Update package dependencies to solve security vulnerabilities shown, when installing `wrtc` via npm.


Dreaded-Gnu:
Fixes #427

markandrus:
Thanks for this. I think there's just one issue: the updated Mocha dependency does not auto exit once the WPT tests complete. Actually, jsdom—which we borrow the WPT test runner from—is [still using the old Mocha version](https://github.com/jsdom/jsdom/blob/master/package.json#L66), too. Newer Mocha releases have a `--exit` flag we can pass via command-line (and probably a programmatic API, too). We should see how to use this in our WPT tests.

markandrus:
I ran with this here: https://github.com/js-platform/node-webrtc/pull/428

markandrus:
Closed via #428

Issue #432, opened 2018-09-07
Unit test 'rtcrtpreceiver.js' crashes


sebas642:
The `rtcrtpreceiver.js` test crashes every single time I run it. The backtrace is the following:
```text
#0  0x00007f7895563f23 in std::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string(std::string const&) ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x00007f7891bc9782 in webrtc::RtpCodecParameters::RtpCodecParameters (this=0x7f78341aa388)
    at ../third_party/webrtc/include/webrtc/api/rtpparameters.h:340
#2  0x00007f7891d4feeb in void std::vector<webrtc::RtpCodecParameters, std::allocator<webrtc::RtpCodecParameters> >::_M_emplace_back_aux<webrtc::RtpCodecParameters>(webrtc::RtpCodecParameters&&) () from /tmp/node-webrtc/build/Debug/wrtc.node
#3  0x00007f7891d476b0 in cricket::WebRtcVoiceMediaChannel::GetRtpReceiveParameters(unsigned int) const ()
   from /tmp/node-webrtc/build/Debug/wrtc.node
#4  0x00007f7891e2aac2 in cricket::VoiceChannel::GetRtpReceiveParameters_w(unsigned int) const () from /tmp/node-webrtc/build/Debug/wrtc.node
#5  0x00007f7891e3524b in rtc::FunctorMessageHandler<webrtc::RtpParameters, rtc::MethodFunctor<cricket::VoiceChannel const, webrtc::RtpParameters (cricket::VoiceChannel::*)(unsigned int) const, webrtc::RtpParameters, unsigned int> >::OnMessage(rtc::Message*) ()
   from /tmp/node-webrtc/build/Debug/wrtc.node
#6  0x00007f7891c97370 in rtc::Thread::ReceiveSendsFromThread(rtc::Thread const*) () from /tmp/node-webrtc/build/Debug/wrtc.node
#7  0x00007f7891dd2807 in rtc::MessageQueue::Get(rtc::Message*, int, bool) () from /tmp/node-webrtc/build/Debug/wrtc.node
#8  0x00007f7891c96f12 in rtc::Thread::Run() () from /tmp/node-webrtc/build/Debug/wrtc.node
[...]

PR #433, opened 2018-09-07
Fix AudioDeviceModule leak

### LINKED ISSUE #429: AudioDeviceModule is leaking
The `AudioDeviceModule` created in the `PeerConnectionFactory` is leaking. That wouldn't be too bad if it was leaking memory only, but it is leaking threads as well and CPU usage is going up over time.

This can be easily observed by running the following:
```javascript
'use strict';

var RTCPeerConnection = require('..').RTCPeerConnection;

setInterval(() => {
  const pc = new RTCPeerConnection();
  pc.close();
  global.gc();
  console.log('## ok');
}, 1000);
```
`$ node --expose-gc test/create-close.js`:

Looking at the process information, we see the number of threads growing over time:
`$ watch "grep Threads /proc/`pgrep -f node.*create-close.js`/status"`

The threads grow twice as fast if `WEBRTC_WIN` is not defined.

sebas642:
Fixes #429 
I've tested only under Linux, with and without `WEBRTC_WIN` defined. The unit tests are passing, except for `rtcrtpreceiver.js` (see issue #432).

I made sure the `AudioDeviceModule` was getting destroyed through the unit tests, but I don't use the mediastreams so I can't actually say that nothing has changed.

### DIFF
--- a/src/peerconnectionfactory.cc
+++ b/src/peerconnectionfactory.cc
@@ -55,15 +55,15 @@ PeerConnectionFactory::PeerConnectionFactory(Maybe<AudioDeviceModule::AudioLayer
   result = _workerThread->Start();
   assert(result);
 
-  auto audioDeviceModule = _workerThread->Invoke<AudioDeviceModule*>(RTC_FROM_HERE, [audioLayer]() {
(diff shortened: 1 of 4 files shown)


markandrus:
The change itself looks good 👍 Did #432 only appear after this change, or could you reproduce before?

sebas642:
#432 was happening before my change, but I don't know when it was introduced.

PR #437, opened 2018-09-13
Restricted HandleStateChange of DataChannel to open and closed

### LINKED ISSUE #436: peer connection close not working properly
Since version `0.2.0` the peer connection close seems to trigger not correctly any longer. When I invoke the `close` method of peer connection, the behaviour is quite a bit strange. After call of `close` the data channel `open` event is triggered. After the data channel `open` event, the `close` event of data channel is triggered.

Tested it also with previous version `0.1.6` and there it doesn't happen. When I invoke the peer connection `close` method with previous version, no data channel `open` event is triggered, and only the data channel `close` event is triggered.

To get around that issue, I've tried to bind the data channel `open` event with a once handler, but that doesn't worked.

After investigating a bit, it seems that the wrong event is emitted after peer connection `close`. Instead of emitting `close`, `open` is emitted. It seams that it's occurring due to the race condition fix of commit [b309a5f66e7441ecd29efe053ddb6b78c7ba7f08](https://github.com/js-platform/node-webrtc/commit/b309a5f66e7441ecd29efe053ddb6b78c7ba7f08). When switching the javascript library back to use `this.readyState` for switch case, `close` is emitted instead of `open`. So there might be something wrong within the native library at `src/datachannel.cc` within `HandleStateEvent`.

Dreaded-Gnu:
Restricted HandleStateChange of DataChannel to open and closed.

Fixes #436

### DIFF
--- a/src/datachannel.cc
+++ b/src/datachannel.cc
@@ -108,11 +108,12 @@ void DataChannel::HandleStateEvent(const DataChannelStateChangeEvent& event) {
   Local<Value> argv[1];
   if (event.state == webrtc::DataChannelInterface::kClosed) {
     argv[0] = Nan::New("closed").ToLocalChecked();
-  } else {
+    MakeCallback("onstatechange", 1, argv);
+  } else if (event.state == webrtc::DataChannelInterface::kOpen) {
     argv[0] = Nan::New("open").ToLocalChecked();
+    MakeCallback("onstatechange", 1, argv);
   }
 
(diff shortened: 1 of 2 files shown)


markandrus:
Thanks for the PR! Would you mind landing a test for #436 that fails before this change and passes after? This would ensure we do not regress in the future.

Dreaded-Gnu:
Added test case failing without change of pull request

markandrus:
Thanks a lot!

PR #450, opened 2018-11-23
Support for cross-compile ARM (Raspberry pi)

ghost:


CMakeLists.txt
```diff
@@ -34,15 +34,31 @@ set(depot_tools_install_dir ${SOURCE_DIR})
...
+else()
+    list(APPEND GN_GEN_ARGS
+      rtc_build_examples=false
+      rtc_build_tools=false
+      rtc_use_x11=false
+      rtc_enable_protobuf=false
+      rtc_use_gtk=false
+      rtc_include_pulse_audio=false
+      rtc_include_tests=false
+  )
```
markandrus, line 57: Can you factor out the common args? For example,

* `rtc_build_examples=false`
* `rtc_use_x11=false`
* `rtc_enable_protobuf=false`
* `rtc_use_gtk=false`
* `rtc_include_pulse_audio=false`
* `rtc_include_tests=false`

CMakeLists.txt
```diff
@@ -219,6 +235,7 @@ endif()
 #
 # is to include all the header files here.
 target_include_directories(${MODULE} SYSTEM PRIVATE
+  #/home/murat/js/node-webrtc-0.3.2/build/external/libwebrtc/download/src/build/linux/debian_sid_arm-sysroot/usr/include/arm-linux-gnueabihf
```
markandrus: Can you remove this?

CMakeLists.txt
```diff
@@ -34,15 +34,31 @@ set(depot_tools_install_dir ${SOURCE_DIR})
...
-list(APPEND GN_GEN_ARGS
-  rtc_build_examples=false
-  rtc_build_tools=false
-  rtc_use_x11=false
-  rtc_enable_protobuf=false
-  rtc_use_gtk=false
-  rtc_include_pulse_audio=false
-  rtc_include_tests=false
-)
+if ("$ENV{TARGET_ARCH}" STREQUAL "arm")
```
markandrus: Rather than use a `TARGET_ARCH` environment variable, should we be using `CMAKE_SYSTEM_PROCESSOR`? Or does this have some sort of interaction with

* [`node-pre-gyp`](https://github.com/mapbox/node-pre-gyp)'s install process
* [`node-cmake`](https://github.com/cjntaylor/node-cmake)'s build process

CMakeLists.txt
```diff
@@ -335,13 +352,39 @@ else()
...
+      set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
+      set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
+      set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
+      set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
+    else()
+      target_compile_options(${MODULE} PRIVATE
+        -B${libwebrtc_source_dir}/src/third_party/binutils/Linux_x64/Release/bin
+        -fpermissive
+        -nostdinc++
+        -nodefaultlibs
```
markandrus: Like above, can you factor out the common args? For example,

* `-fpermissive`
* `-nostdinc++`
* `-nodefaultlibs`

CMakeLists.txt
markandrus: I guess `TARGET_ARCH` will work better with `npm i`?

scripts/configure-webrtc.sh
```diff
@@ -8,4 +8,10 @@ export PATH=$DEPOT_TOOLS:$PATH
 
 cd ${SOURCE_DIR}
 
+if [ "$TARGET_ARCH" == "arm" ]; then
+    python build/linux/sysroot_scripts/install-sysroot.py --arch=arm
+else
+    python build/linux/sysroot_scripts/install-sysroot.py --arch=amd64
```
markandrus, line 14: Looks like this was not called before. Is it actually necessary for the non-ARM case?

markandrus: Actually, it looks like a similar line is in `scripts/download-webrtc.sh`. Should both of these move there?

markandrus (changes requested):
@doganmurat thanks a lot for this. It's looking good. I had some initial comments, plus some asks:

1. Can you add entries to `.travis.yml` and/or `.circleci/config.yml` to actually do the cross-compilation in CI?
2. How can we test this in CI?
3. Will downloading pre-compiled binaries work when setting `TARGET_ARCH`? For example, if you want to build an app for ARM on a non-ARM machine, then deploy to your particular device. Or does ARM deployment not typically work this way (i.e., just do the `npm i` on the ARM device itself)?

CMakeLists.txt
ghost: OK, it could be better.
I will make it.

CMakeLists.txt
ghost: OK

CMakeLists.txt
ghost: I think TARGET_ARCH will work better.
We will have more control over process.

CMakeLists.txt
ghost: OK

scripts/configure-webrtc.sh
ghost: > Looks like this was not called before. Is it actually necessary for the non-ARM case?

It was called from download-webrtc.sh, but we should call this from configure-webrtc.sh.
If you run "TARGET_ARCH=arm npm i" second time, process will not call download-webrtc.sh because it is already downloaded.

ghost: > Actually, it looks like a similar line is in `scripts/download-webrtc.sh`. Should both of these move there?

It is moved to here.

CMakeLists.txt
ghost: - I don't have any experience for travis, but it looks straight.
I will try to make required modifications, but it could be lots of failed builds :)

- The only test I can think is get binary file info & check ELF version.  It should look somtething like that
`
build/Release/wrtc.node: ELF 32-bit LSB shared object, ARM, EABI5 version 1 (GNU/Linux), dynamically linked, BuildID[sha1]=a3a65b00d8e983bc282f8ddf9db80986fe8b69bd, not stripped
` 

- Will downloading pre-compiled binaries work when setting TARGET_ARCH? No
If you want to compile it from scratch then you will use these arguments. Normally if we have binaries on s3 server then user will run a "npm i" and pre-compiled binaries will be downloaded.
Of course if you are on an ARM platform :)

ghost: Ok it is not so straight, I need help for travis.
Could you please help it?
Here is necessary commands for cross-compiling.
```
wget https://releases.linaro.org/components/toolchain/binaries/latest-7/arm-linux-gnueabihf/gcc-linaro-7.3.1-2018.05-i686_arm-linux-gnueabihf.tar.xz
tar xf gcc-linaro-7.3.1-2018.05-i686_arm-linux-gnueabihf.tar.xz
SKIP_DOWNLOAD=true TARGET_ARCH=arm ARM_TOOLS_PATH=$(pwd)/gcc-linaro-7.3.1-2018.05-i686_arm-linux-gnueabihf npm install
```

markandrus: Thanks for answering my questions, and yes, I will try it. Thanks!

CMakeLists.txt
markandrus: Sounds good 👍

markandrus:
Thanks for making those changes, @doganmurat. I'm working on getting Travis setup to cross-compile over on [arm-support](https://github.com/js-platform/node-webrtc/tree/arm-support) (I'll open a PR once it's working), but I'm hitting the following error from `/usr/bin/ld.gold`:

```
[1922/2221] LINK ./rgba_to_i420_converter
FAILED: rgba_to_i420_converter 
python "../../download/src/build/toolchain/gcc_link_wrapper.py" --output="./rgba_to_i420_converter" -- ../../download/src/third_party/llvm-build/Release+Asserts/bin/clang++ -fPIC -Wl,-z,noexecstack -Wl,-z,now -Wl,-z,relro -Wl,-z,defs -Wl,--as-needed -fuse-ld=gold -B../../download/src/third_party/binutils/Linux_x64/Release/bin -Wl,--icf=all --target=arm-linux-gnueabihf -Wl,--gdb-index -rdynamic -nostdlib++ --sysroot=../../download/src/build/linux/debian_sid_arm-sysroot -L../../download/src/build/linux/debian_sid_arm-sysroot/usr/local/lib/arm-linux-gnueabihf -Wl,-rpath-link=../../download/src/build/linux/debian_sid_arm-sysroot/usr/local/lib/arm-linux-gnueabihf -L../../download/src/build/linux/debian_sid_arm-sysroot/lib/arm-linux-gnueabihf -Wl,-rpath-link=../../download/src/build/linux/debian_sid_arm-sysroot/lib/arm-linux-gnueabihf -L../../download/src/build/linux/debian_sid_arm-sysroot/usr/lib/arm-linux-gnueabihf -Wl,-rpath-link=../../download/src/build/linux/debian_sid_arm-sysroot/usr/lib/arm-linux-gnueabihf -Wl,-rpath-link=. -Wl,--disable-new-dtags -o "./rgba_to_i420_converter" -Wl,--start-group @"./rgba_to_i420_converter.rsp"  -Wl,--end-group   -ldl -lpthread -lrt 
/usr/bin/ld.gold: fatal error: ../../download/src/build/linux/debian_sid_arm-sysroot/usr/lib/gcc/arm-linux-gnueabihf/6/../../../arm-linux-gnueabihf/crt1.o: unsupported ELF machine number 40
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Off the top of my head, I don't think `/usr/bin/ld.gold` should be used, since it's not the ARM version;  but I'm not quite sure what's up yet.

markandrus:
After installing `g++-arm-linux-gnueabihf` in Travis, the build apparently succeeds; however, I’d like to test the resulting binary. Actually, it may make sense to duplicate these results in CircleCI, since they may have better support for spinning up an ARM container to test the artifact. Or maybe we can test another way.

ghost:
These are very good news.
I am using wrtc  since v0.0.5x and I am compiling all binaries for arm myself.
I am testing arm binary in a qemu image and also in a real raspberry pi.
I don’t have much info about travis so I couldn’t help about how to do it in travis.
If you need help that I can do pls inform me.

Best regards 
Murat

doganmurat:
By the way,
You will except my PR right?

PR #455, opened 2018-11-26
Add install-build-deps command

ghost:
In order to install all required packages to build webrtc locally we should call src/build/install-build-deps.sh script.

### DIFF
--- a/scripts/download-webrtc.sh
+++ b/scripts/download-webrtc.sh
@@ -16,6 +16,8 @@ else
   download_from_google_storage --no_resume --platform=${PLATFORM} --no_auth --bucket chromium-gn -s src/buildtools/linux64/gn.sha1
 fi
 
+sudo src/build/install-build-deps.sh
+
 python src/tools/clang/scripts/update.py
 
 rm -f webrtc


ghost:
At travis platform this is not a problem but for a local compilation this script must be called before compilation. Without calling this script build will fail.

markandrus:
Thanks for the PR, @doganmurat, but I don't think we should land this as-is for a couple of reasons:

1. `npm install` should not mutate a user's system globally, i.e. we should not call `sudo` or install things outside of the node-webrtc directory. There was similar feedback from [another user](https://github.com/node-webrtc/node-webrtc/issues/448#issuecomment-440304237).
2. There are already things a user must do to build from source that node-webrtc's build system doesn't take care of. For examples, installing a C++ compiler, installing an appropriate version of CMake, etc. IMO, it would be better to treat those build dependencies installed by install-build-deps.sh similarly.

That being said, are there reasons you need such a change?

ghost:
I didn’t think in that way, you could be right.
But in most cases user should call this script before a successful compilation.
What about printing some notification messages?

markandrus:
> What about printing some notification messages?

If you think it would be useful, please consider another PR. I think it would be good to test for the necessary tool(s) in order to determine whether to print the notification.

PR #459, opened 2018-12-10
Support arm64/aarch64 native build

zsmweb:
wget https://releases.linaro.org/components/toolchain/binaries/latest-7/aarch64-linux-gnu/gcc-linaro-7.3.1-2018.05-x86_64_aarch64-linux-gnu.tar.xz
tar xf gcc-linaro-7.3.1-2018.05-x86_64_aarch64-linux-gnu.tar.xz
SKIP_DOWNLOAD=true TARGET_ARCH=arm64 ARM_TOOLS_PATH=$(pwd)/gcc-linaro-7.3.1-2018.05-x86_64_aarch64-linux-gnu npm install

PR #460, opened 2018-12-14
arm64 tweaks

markandrus:
Followup on #459


zsmweb:
You're so generous! Thanks!
Sorry，I'm busy with other things for the time being.
But I believe if you can do better than I can.

markandrus:
NP, @zsmweb. If you don't mind testing the binaries once published, that would be great 👍

PR #466, opened 2018-12-23
Use X Macros to simplify (?) converters

markandrus:
OK, I'm not sure if this was a great idea or not, but I started trying to simplify the code that converts between C++ and JavaScript. I do this using a combination of [X Macros](https://en.wikipedia.org/wiki/X_Macro) and the syntax introduced to support `fmap` and `ap` on [applicative functors](https://en.wikipedia.org/wiki/Applicative_functor) like `node_webrtc::Validation`.

In Haskell, there are great libraries like [optparse-applicative](https://github.com/pcapriotti/optparse-applicative) that let you build up a parser for a dictionary type if you have subparsers for each field of the dictionary. I wanted to achieve something similar. The result is somewhat perverse, and I'm not yet sure if simpler.


markandrus:
I think I will split up `converters/webrtc.{cc,h}` into `converters/enums.{cc,h}` and `converters/dictionaries.{cc,h}`.

markandrus:
Next, I'll split `src/converters/interfaces.h` off of `src/converters/dictionaries.h`.

PR #472, opened 2019-01-08
Programmatic Video

markandrus:
The idea here is to bind `rtc::AdaptedVideoTrackSource` and expose it as `RTCVideoSource`. `RTCVideoSource` will likely have, at a minimum, the following interface:

```webidl
[constructor(optional RTCVideoSourceConfiguration configuration)]
interface RTCVideoSource {
  MediaStreamTrack createTrack();
};

dictionary RTCVideoSourceConfiguration {
  int width;
  int height;
};
```

An `RTCVideoSource` would allow creating N `MediaStreamTrack`s. The next question is whether to expose a "push" or "pull" model for sending frames. For "push", maybe we just want the user to call a method, `onFrame`, as quickly as  they like:

```webidl
partial interface RTCVideoSource {
  void onFrame(RTCVideoFrame frame);  // *
};
```

For "pull", maybe we want the library to read from some buffer, automatically, at a specified interval. Then we probably need a `stop` method and to augment `RTCVideoSourceConfiguration`, too. **

```webidl
partial interface RTCVideoSource {
  void stop();
};

partial dictionary RTCVideoSourceConfiguration {
  int frameRate;
  Uint8ClampedArray data;  // *
};
```

\* There is a DOM API, [`ImageData`](https://developer.mozilla.org/en-US/docs/Web/API/ImageData), that might be close to what we need for representing a frame's underlying `webrtc::I420Buffer`. `ImageData` has a `data` member of type `Uint8ClampedArray`.  We could use one or the other.
\** "pull" feels closer to an existing [`captureStream`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/captureStream) DOM API.

Questions:

- [x] ~~"push" or "pull" API~~ "push" API seems fine for now.
- [x] ~~Allow changing dimensions on the fly.~~ Works with the "push" API.
- [x] ~~What about reading from files? This could be done in JavaScript, but a native source might work better.~~ Likely true, but can be tackled in a separate PR.
- [x] ~~How should a renderer API for getting frames out look?~~ See RTCVideoSink below.

---

**Update 1:** I've added an example of using `RTCVideoSource` for programmatic video generation. You can run it as follows:

```
node examples/programmatic-video.server.js
```

then navigate to [http://localhost:8080/programmatic-video.client.html](http://localhost:8080/programmatic-video.client.html). You'll receive a remote video `MediaStreamTrack` whose frames are generated by the Node server. The example includes a simple animation (pictured below), but it would be fun to hook this up with a project like [node-canvas](https://github.com/Automattic/node-canvas).

<image src="https://user-images.githubusercontent.com/359154/50874829-2abe3080-138b-11e9-8dd7-c229e9ae825a.gif" width="160" height="120" />

---

**Update 2:** I've updated examples/programmatic-video.server.js to use node-canvas. I also landed a rough JavaScript implementation for converting ARGB32 to I420; although, it would probably be better to bind to libyuv.

<image src="https://user-images.githubusercontent.com/359154/51068381-056f3380-15d2-11e9-86c9-11f3823248d3.gif" width="160" />

---

**Update 3:** I've added `RTCVideoSink` bindings to expose `rtc::VideoSinkInterface<webrtc::VideoFrame>`. The API is basically

```webidl
[constructor(MediaStreamTrack track)]
interface RTCVideoSink {
  attribute EventHandler onframe;
  void stop();
};

dictionary RTCVideoFrame {
  int width;
  int height;
  Uint8ClampedArray data;
};
```

although I think we should have a proper `EventTarget`.

---

**Update 4:** I've updated examples/programmatic-video.server.js to use node-canvas and `RTCVideoSink` to show some simple video compositing.

---

**Update 5:** I added some latency tests (`test/latency.js`) which measure the time between pushing a video frame through `RTCVideoSource` to receiving it from `RTCVideoSink`. These—and some other things—still need some work, though. For example,

- [x] The latency tests: Not all syntax is supported in Node 6, and I'd rather not rewrite it. So we should not run these in Node 6 (which will also EOL in a couple months anyway).
- [x] Non-standard API additions: I'm thinking non-standard objects like RTCVideoSource and RTCVideoSink should live under a `nonstandard` prefix. For example,

    ```js
    const { RTCVideoSource, RTCVideoSink } = require('wrtc').nonstandard;
    ```
- [x] ~~`EventTarget` interface for `RTCVideoSink`.~~ I landed this in JavaScript via `lib/rtcvideosink.js` so that we could get this out sooner; I'd rather figure out how to do this properly in C++, though.
- [x] Review usage of stride, or lack thereof: when passing frames to `RTCVideoSource`, I think we can define some defaulting behavior for stride, if even necessary (actually, I'd rather just punt on user-defined stride for user-provided frames); however, when receiving frames from `RTCVideoSink`, I think we must consider it, whereas we don't currently.
- [x] ~~Should we have a proper interface for `RTCVideoFrame`, or is a plain object sufficient? In the interest of landing this sooner than later, and getting some feedback/enabling use cases, probably the object is fine.~~ Plain object should be fine for this PR.
- [x] ~~Should we have more libyuv bindings? Yes, probably.~~ Done. We now have `i420ToRgba` and `rgbaToI420`.

---

**Update 6:** I think it's basically ready. I'll look at releasing in the next few days after another review.


CHANGELOG.md
```diff
@@ -4,6 +4,115 @@
...
+### Programmatic Video
+
+This release of node-webrtc adds non-standard, programmatic video APIs in the
+form of RTCVideoSource and RTCVideoSink. With these APIs, you can
+
+* Pass [I420](https://wiki.videolan.org/YUV/#I420) frames to RTCVideoSource via
+  the `onFrame` method. Then use RTCVideoSource's `createTrack` method to create
+  a local video MediaStreamTrack.
+* Construct an RTCVideoSink from a local or remote video MediaStreamTrack. The
+  RTCVideoSink will emit a "frame" event everytime an I420 frame is received.
```
markandrus: "everyime" → "every time"

CHANGELOG.md
```diff
@@ -4,6 +4,115 @@
...
+  When you're finished, stop the RTCVideoSink by calling `stop`.
```
markandrus, line 17: So, I think it makes sense to have a `stop` method on the RTCVideoSink, because an RTCVideoSink should be allowed to stop before its MediaStreamTrack does; however, what if the MediaStreamTrack stops?

- [x] ~~Shouldn't the RTCVideoSink be stopped automatically?~~
- [x] Also, should we have some introspection on the stopped state, i.e.`stopped`?

lib/index.js
```diff
@@ -14,3 +14,9 @@ exports.RTCRtpReceiver = binding.RTCRtpReceiver;
 exports.RTCRtpSender = binding.RTCRtpSender;
 exports.RTCRtpTransceiver = binding.RTCRtpTransceiver;
 exports.RTCSessionDescription = require('./sessiondescription');
+
+exports.nonstandard = {};
+exports.nonstandard.RTCVideoSink = require('./rtcvideosink');
+exports.nonstandard.RTCVideoSource = binding.RTCVideoSource;
+exports.nonstandard.rgbaToI420 = binding.rgbaToI420;
+exports.nonstandard.i420ToRgba = binding.i420ToRgba;
```
markandrus: Alphabetize the `nonstandard` APIs?

src/binding.cc
```diff
@@ -11,13 +11,16 @@
...
+#include "src/i420helpers.h"
 #include "src/mediastream.h"  // IWYU pragma: keep
 #include "src/mediastreamtrack.h"  // IWYU pragma: keep
 #include "src/rtcrtpreceiver.h"  // IWYU pragma: keep
 #include "src/rtcrtpsender.h"  // IWYU pragma: keep
 #include "src/rtcrtptransceiver.h"  // IWYU pragma: keep
 #include "src/legacyrtcstatsreport.h"  // IWYU pragma: keep
 #include "src/rtcstatsresponse.h"  // IWYU pragma: keep
+#include "src/rtcvideosource.h"
+#include "src/rtcvideosink.h"
```
markandrus: Alphabetize?

src/converters/dictionaries.cc
```diff
@@ -643,6 +643,34 @@ TO_JS_IMPL(rtc::scoped_refptr<webrtc::RTCStatsReport>, value) {
...
+  auto sizeOfLuminancePlane = width * height;
+  auto sizeOfChromaPlane = sizeOfLuminancePlane / 4;
+
+  auto actualByteLength = static_cast<int>(contents.ByteLength());
+  auto expectedByteLength = sizeOfLuminancePlane + sizeOfChromaPlane + sizeOfChromaPlane;
+  if (actualByteLength != expectedByteLength) {
+    return node_webrtc::Validation<rtc::scoped_refptr<webrtc::I420Buffer>>::Invalid(
+            "Expected a .byteLength of " + std::to_string(expectedByteLength) + ", not " +
+            std::to_string(actualByteLength));
+  }
```
markandrus: There is overlap here with the converter for I420ImageData. These two should be consolidated and we should have a

```cpp
CONVERT_VIA(v8::Local<v8::Value>, I420ImageData, rtc::scoped_refptr<webrtc::I420Buffer>)
```

This also means that I420ImageData (and the related classes) need to be defined in `src/dictionaries.h`.

src/converters/dictionaries.cc
```diff
@@ -643,6 +643,34 @@ TO_JS_IMPL(rtc::scoped_refptr<webrtc::RTCStatsReport>, value) {
...
+
+  uint8_t* sourceDataY = static_cast<uint8_t*>(contents.Data());
+  uint8_t* sourceDataU = sourceDataY + sizeOfLuminancePlane;
+  uint8_t* sourceDataV = sourceDataU + sizeOfChromaPlane;
```
markandrus: If we implement what is described above, then it would be useful to have pointers to these fields readily available on I420ImageData.

src/converters/dictionaries.cc
```diff
@@ -656,6 +684,64 @@ OBJ_FROM_JS_IMPL1(RTCSESSIONDESCRIPTIONINIT, CreateRTCSessionDescriptionInit)
...
 #undef DEFAULT
+
+// TODO(mroberts): We should have OBJ_TO_JS_IMPL for such simple cases.
+TO_JS_IMPL(webrtc::VideoTrackSourceInterface::Stats, value) {
+  Nan::EscapableHandleScope scope;
+  auto stats = Nan::New<v8::Object>();
+  stats->Set(Nan::New("width").ToLocalChecked(), node_webrtc::From<v8::Local<v8::Value>>(value.input_width).UnsafeFromValid());
+  stats->Set(Nan::New("height").ToLocalChecked(), node_webrtc::From<v8::Local<v8::Value>>(value.input_height).UnsafeFromValid());
+  return node_webrtc::Pure(scope.Escape(stats).As<v8::Value>());
+}
```
markandrus: I don't think `webrtc::VideoTrackSourceInterface::Stats` was actually used, so we could delete this.

src/rtcvideosource.cc
```diff
@@ -0,0 +1,99 @@
...
+
+NAN_GETTER(node_webrtc::RTCVideoSource::GetIsScreencast) {
+  auto self = Nan::ObjectWrap::Unwrap<node_webrtc::RTCVideoSource>(info.Holder());
+  info.GetReturnValue().Set(self->_source->is_screencast());
+}
+
+NAN_GETTER(node_webrtc::RTCVideoSource::GetRemote) {
+  auto self = Nan::ObjectWrap::Unwrap<node_webrtc::RTCVideoSource>(info.Holder());
+  info.GetReturnValue().Set(self->_source->remote());
+}
```
markandrus: Allow configuring

- [x] `needsDenoising`
- [x] `isScreencast`
- [x] ~~`remote`~~

at construction time via some RTCVideoSourceInit.

markandrus: Actually, I can see no reason to override `remote`, so scratch that one.

src/rtcvideosource.cc
```diff
@@ -0,0 +1,99 @@
...
+
+NAN_GETTER(node_webrtc::RTCVideoSource::GetState) {
+  auto self = Nan::ObjectWrap::Unwrap<node_webrtc::RTCVideoSource>(info.Holder());
+  auto state = node_webrtc::From<v8::Local<v8::Value>>(self->_source->state());
+  info.GetReturnValue().Set(state.UnsafeFromValid());
+}
```
markandrus: This is always going to be "live", so should we just remove it and the converter?

src/rtcvideosink.cc
```diff
@@ -0,0 +1,72 @@
...
+ * project authors may be found in the AUTHORS file in the root of the source
+ * tree.
+ */
+#include "src/rtcvideosink.h"
+
+#include "src/converters.h"
+#include "src/converters/arguments.h"
+#include "src/converters/dictionaries.h"
+#include "src/converters/interfaces.h"
+#include "peerconnectionfactory.h"
```
markandrus: Should be prefixed with "src/".

src/rtcvideosink.cc
```diff
@@ -0,0 +1,72 @@
...
+
+node_webrtc::RTCVideoSink::RTCVideoSink(rtc::scoped_refptr<webrtc::VideoTrackInterface> track)
+  : node_webrtc::AsyncObjectWrapWithLoop<node_webrtc::RTCVideoSink>("RTCVideoSink", *this)
+  , _track(std::move(track)) {
+  rtc::VideoSinkWants wants;
+  _track->AddOrUpdateSink(this, wants);
+  // node_webrtc::PeerConnectionFactory::GetOrCreateDefault()->_signalingThread->Invoke<void>(RTC_FROM_HERE, [this]() {
+  //   rtc::VideoSinkWants wants;
+  //   this->_track->AddOrUpdateSink(this, wants);
+  // });
```
markandrus: Delete this.

src/rtcvideosink.cc
```diff
@@ -0,0 +1,72 @@
...
+  info.GetReturnValue().Set(info.This());
+}
+
+NAN_METHOD(node_webrtc::RTCVideoSink::JsStop) {
+  auto self = AsyncObjectWrapWithLoop<node_webrtc::RTCVideoSink>::Unwrap(info.Holder());
+  self->Stop();
+  if (self->_track) {
+    self->_track->RemoveSink(self);
+    self->_track = nullptr;
+  }
```
markandrus: Now that we can override `Stop`, this could actually move in there.

src/rtcvideosource.cc
```diff
@@ -0,0 +1,99 @@
...
+#include <webrtc/media/base/videocapturer.h>
+
+#include "src/converters.h"
+#include "src/converters/arguments.h"
+#include "src/converters/v8.h"
+#include "src/converters/dictionaries.h"
+#include "src/functional/maybe.h"
+#include "src/functional/validation.h"
+#include "src/mediastreamtrack.h"
+#include "src/peerconnectionfactory.h"
```
markandrus: Alphabetize.

src/rtcvideosink.cc
```diff
@@ -0,0 +1,72 @@
...
+  //   this->_track->AddOrUpdateSink(this, wants);
+  // });
+}
+
+NAN_METHOD(node_webrtc::RTCVideoSink::New) {
+  CONVERT_ARGS_OR_THROW_AND_RETURN(track, rtc::scoped_refptr<webrtc::VideoTrackInterface>);
+  auto sink = new RTCVideoSink(track);
+  sink->Wrap(info.This());
+  info.GetReturnValue().Set(info.This());
+}
```
markandrus, line 39: This needs the "new" check (see RTCVideoSource, for example).

src/rtcvideosource.cc
```diff
@@ -0,0 +1,99 @@
...
+
+  info.GetReturnValue().Set(result->ToObject());
+}
+
+NAN_METHOD(node_webrtc::RTCVideoSource::OnFrame) {
+  auto self = Nan::ObjectWrap::Unwrap<node_webrtc::RTCVideoSource>(info.Holder());
+  CONVERT_ARGS_OR_THROW_AND_RETURN(buffer, rtc::scoped_refptr<webrtc::I420Buffer>);
+  webrtc::VideoFrame::Builder builder;
+  auto frame = builder.set_video_frame_buffer(buffer).build();
+  self->_source->PushFrame(frame);
```
markandrus, line 66: I think this could move into a converter, but I'm not sure if it's worth it.

test/i420helpers.js
```diff
@@ -0,0 +1,40 @@
...
+    const rgbaData = new Uint8ClampedArray(width * height * 4);
+    const rgbaFrame = { width, height, data: rgbaData };
+
+    blackenI420Frame(i420Frame);
+    i420ToRgba(i420Frame, rgbaFrame);
+    // TODO(mroberts): Test the result.
+
+    whitenI420Frame(i420Frame);
+    i420ToRgba(i420Frame, rgbaFrame);
+    // TODO(mroberts): Test the result.
```
markandrus: We really ought to test this, because it would be good to know if there's any endian-weirdness on certain platforms.

test/lib/bits.js
```diff
@@ -0,0 +1,281 @@
+'use strict';
```
markandrus, line 1: The latency test only uses `fromBits` and `toBits` now, so clean this up.

test/lib/timestamp.js
```diff
@@ -0,0 +1,253 @@
+'use strict';
```
markandrus, line 1: The latency test only uses `printBitsI420` and `readBitsI420` now, so clean this up.

markandrus (commented):
Notes to self&hellip;

CHANGELOG.md
markandrus: Done.

lib/index.js
markandrus: Done.

src/binding.cc
markandrus: Done.

src/converters/dictionaries.cc
markandrus: Done.

src/rtcvideosource.cc
markandrus: Done.

src/rtcvideosink.cc
markandrus: Done.

src/rtcvideosink.cc
markandrus: Done.

src/rtcvideosink.cc
markandrus: Done.

src/rtcvideosource.cc
markandrus: Done.

src/rtcvideosink.cc
markandrus: Done.

test/lib/bits.js
markandrus: Done.

test/lib/timestamp.js
markandrus: Done.

test/i420helpers.js
markandrus: Done.

src/rtcvideosource.cc
markandrus: Done. I also removed `remote`'s getter and documented RTCVideoSourceInit.

src/rtcvideosource.cc
markandrus: Skip.

CHANGELOG.md
markandrus: Regarding stopping the RTCVideoSink automatically, unfortunately we cannot just use the `webrtc::ObserverInterface`, since this appears to act asynchronously. Instead, if we wanted to do this, we need to build our own machinery for letting the MediaStreamTrack tell the RTCVideoSink it stopped, which is a bit onerous&hellip; Also, automatically stopping would preclude re-use of RTCVideoSinks with other MediaStreamTracks, so maybe we don't want to do this anyway.

markandrus: Done.

src/converters/dictionaries.cc
markandrus: Done.

src/converters/dictionaries.cc
markandrus: Done.

mturnshek:
This is awesome. Your efforts are appreciated here! Great examples too.

PR #473, opened 2019-01-09
Enable Electron build

alessiopcc:
This PR allow to build this module for Electron using npm.
As described in [https://electronjs.org/docs/tutorial/using-native-node-modules#using-npm](https://electronjs.org/docs/tutorial/using-native-node-modules#using-npm) just set a few environment variables before installing:
```
npm_config_runtime=electron
npm_config_target=<electron version>
npm_config_disturl=https://atom.io/download/electron
```

Tested with Electron 4
Successfully built for Electron 2+


markandrus:
Thanks a lot, @alessiopcc. I will merge this. Do you think we should ship binaries for Electron?

alessiopcc:
@markandrus 
I'm using it, so it would be great to have prebuilt of course. I'm a bit biased I think.
In case you decide to ship it, I would suggest to support Electron 3.1 and maybe 3.0 (node-pre-gyp use major and minor numbers) that are quite new.
Electron 4.0 works fine on Linux but I've found a crashing bug on Windows platform that I think is Electron related, so it would be not completely supported.

jsherwani:
~@alessiopcc I just tried using the above instructions to compile for Electron 4.1.3 (released just now) on Mac, and end up getting the following error:~

```
Error: Cannot open [...]/wrtc.node: Error: The module '[...]/wrtc.node'
was compiled against a different Node.js version using
NODE_MODULE_VERSION 64. This version of Node.js requires
NODE_MODULE_VERSION 69. Please try re-compiling or re-installing
the module (for instance, using `npm rebuild` or `npm install`).
```

I got it working by following your steps above and running `npm i wrtc` instead of trying to compile it for Electron standalone and using `npm link`. Leaving this here in case someone else has the same issue.

Issue #475, opened 2019-01-14
Add support for setting MediaStreamTrack.enabled


taylorhoward92:
I need the ability to mute peers from the server side. Imagine an audio bridge where only one peer is allowed to be heard at a time. The [recommended way to do this ](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/enabled)would be to use: `MediaStreamTrack.enabled = false`. Currently setting this property has no effect (read-only). Is support for setting this property currently being worked on? Would it be a large piece of work to implement?

markandrus:
I think it should not be hard, but I haven't looked into it. Maybe we can do it for the next release

taylorhoward92:
What are the guidelines for contributing to this repo? If you could point be in the right direction in regards to how this would work at a high level I can have a look into implementing it.

markandrus:
I recommend first

1. Checkout the code.
2. Ensure you can build it.
3. Run the tests. Validate that they pass.

If that works, you should be OK to start making changes. I do my development with [CLion](https://www.jetbrains.com/clion/)—which has good CMake support—and vim, but you could use anything you like. C++ code is in `src/`, JavaScript code is in `lib/`, and tests are in `test/`. There should be existing examples of setters in other C++ files you can copy. I think, roughly, you'll want to

1. Hook up the setter in `mediastreamtrack.cc`. It doesn't have to do anything yet.
2. Identify the API on `webrtc::MediaStreamTrackInterface` (or whatever the appropriate class is) for updating the `enabled` property.
3. Invoke that API in your setter. Sometimes you have to invoke WebRTC APIs on particular threads. If you invoke it on the wrong thread, the debug build will let you know (by failing an `assert`).
4. Add a test for `enabled`. For example, if you set `enabled`, getting `enabled` should return the set value.

I think Step 2 is really the mysterious part and the reason we don't have that feature yet. I think I looked into it briefly but didn't know which API to invoke. Also, my development cycle is basically,

1. Edit the code.
2. Run `./node_modules/.bin/ncmake build` (or let CLion do this for me).
3. Run the test I'm working on (or let CLion's debug feature do this for me).

taylorhoward92:
Thanks for the tips. I've implemented the functionality and created a test as suggested. Code is building and tests are passing. I also did a functional test by disabling the MediaStreamTrack in the loopback server example to confirm black pixels and no audio. Are you able to add me as a contributor so I can push my branch and create a PR?

[...]

PR #476, opened 2019-01-16
Added ability to set .enabled on MediaStreamTrack

taylorhoward92:
See #475. Useful for mute-like functionality [as per the WebRTC API](https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/enabled).

### DIFF
--- a/src/mediastreamtrack.cc
+++ b/src/mediastreamtrack.cc
@@ -80,6 +80,16 @@ NAN_GETTER(MediaStreamTrack::GetEnabled) {
   info.GetReturnValue().Set(Nan::New(self->_track->enabled()));
 }
 
+NAN_SETTER(MediaStreamTrack::SetEnabled) {
+  (void) property;
+
+  auto self = AsyncObjectWrapWithLoop<MediaStreamTrack>::Unwrap(info.Holder());
+
+  CONVERT_OR_THROW_AND_RETURN(value, enabled, boolean_t);
+
+  self->_track->set_enabled(enabled);
+}
+
 NAN_GETTER(MediaStreamTrack::GetId) {
   (void) property;
   auto self = AsyncObjectWrapWithLoop<MediaStreamTrack>::Unwrap(info.Holder());
@@ -177,7 +187,7 @@ void MediaStreamTrack::Init(Handle<Object> exports) {
   MediaStreamTrack::tpl().Reset(tpl);
   tpl->SetClassName(Nan::New("MediaStreamTrack").ToLocalChecked());
   tpl->InstanceTemplate()->SetInternalFieldCount(1);
-  Nan::SetAccessor(tpl->InstanceTemplate(), Nan::New("enabled").ToLocalChecked(), GetEnabled, nullptr);
+  Nan::SetAccessor(tpl->InstanceTemplate(), Nan::New("enabled").ToLocalChecked(), GetEnabled, SetEnabled);
   Nan::SetAccessor(tpl->InstanceTemplate(), Nan::New("id").ToLocalChecked(), GetId, nullptr);
   Nan::SetAccessor(tpl->InstanceTemplate(), Nan::New("kind").ToLocalChecked(), GetKind, nullptr);
   Nan::SetAccessor(tpl->InstanceTemplate(), Nan::New("readyState").ToLocalChecked(), GetReadyState, nullptr);
--- a/src/mediastreamtrack.cc
+++ b/src/mediastreamtrack.cc
@@ -64,6 +64,12 @@ NAN_METHOD(MediaStreamTrack::New) {
   info.GetReturnValue().Set(info.This());
 }
 
+void MediaStreamTrack::Stop() {
+  _ended = true;
+  _enabled = _track->enabled();
+  node_webrtc::AsyncObjectWrapWithLoop<node_webrtc::MediaStreamTrack>::Stop();
+}
+
 void MediaStreamTrack::OnChanged() {
   if (this->_track->state() == webrtc::MediaStreamTrackInterface::TrackState::kEnded) {
     Stop();
@@ -77,7 +83,7 @@ void MediaStreamTrack::OnPeerConnectionClosed() {
 NAN_GETTER(MediaStreamTrack::GetEnabled) {
   (void) property;
   auto self = AsyncObjectWrapWithLoop<MediaStreamTrack>::Unwrap(info.Holder());
-  info.GetReturnValue().Set(Nan::New(self->_track->enabled()));
+  info.GetReturnValue().Set(Nan::New(self->_ended ? self->_enabled : self->_track->enabled()));
 }
 
 NAN_SETTER(MediaStreamTrack::SetEnabled) {
@@ -87,7 +93,11 @@ NAN_SETTER(MediaStreamTrack::SetEnabled) {
 
   CONVERT_OR_THROW_AND_RETURN(value, enabled, bool);
 
-  self->_track->set_enabled(enabled);
+  if (self->_ended) {
+    self->_enabled = enabled;
+  } else {
+    self->_track->set_enabled(enabled);
+  }
 }
 
 NAN_GETTER(MediaStreamTrack::GetId) {
@@ -145,8 +155,6 @@ NAN_METHOD(MediaStreamTrack::Clone) {
   }
   auto clonedMediaStreamTrack = wrap()->GetOrCreate(self->_factory, clonedTrack);
   if (self->_ended) {
-    // NOTE(mroberts): There ought to be a more principled way of doing this; also, we aren't considering the source.
-    clonedMediaStreamTrack->_ended = true;
     clonedMediaStreamTrack->Stop();
   }
   info.GetReturnValue().Set(clonedMediaStreamTrack->ToObject());
@@ -154,7 +162,6 @@ NAN_METHOD(MediaStreamTrack::Clone) {
 
 NAN_METHOD(MediaStreamTrack::JsStop) {
   auto self = AsyncObjectWrapWithLoop<MediaStreamTrack>::Unwrap(info.Holder());
-  self->_ended = true;
   self->Stop();
 }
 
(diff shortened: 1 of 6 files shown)


src/mediastreamtrack.cc
```diff
@@ -80,6 +80,16 @@ NAN_GETTER(MediaStreamTrack::GetEnabled) {
   info.GetReturnValue().Set(Nan::New(self->_track->enabled()));
 }
 
+NAN_SETTER(MediaStreamTrack::SetEnabled) {
+  (void) property;
+
+  auto self = AsyncObjectWrapWithLoop<MediaStreamTrack>::Unwrap(info.Holder());
+
+  CONVERT_OR_THROW_AND_RETURN(value, enabled, boolean_t);
```
markandrus: I think you need `bool` rather than `boolean_t`. We've defined the converter for `bool` [here](https://github.com/node-webrtc/node-webrtc/blob/develop/src/converters/v8.h#L66).

markandrus (dismissed):
Thanks, looking good. Builds fail due to the converter, AFAICT.

```
In file included from /root/node-webrtc/src/events.h:20:0,
                 from /root/node-webrtc/src/eventqueue.h:16,
                 from /root/node-webrtc/src/eventloop.h:17,
                 from /root/node-webrtc/src/promisefulfillingeventloop.h:17,
                 from /root/node-webrtc/src/asyncobjectwrapwithloop.h:17,
                 from /root/node-webrtc/src/mediastreamtrack.h:18,
                 from /root/node-webrtc/src/mediastreamtrack.cc:8:
/root/node-webrtc/src/mediastreamtrack.cc: In static member function 'static Nan::NAN_SETTER_RETURN_TYPE node_webrtc::MediaStreamTrack::SetEnabled(v8::Local<v8::String>, v8::Local<v8::Value>, Nan::NAN_SETTER_ARGS_TYPE)':
/root/node-webrtc/src/mediastreamtrack.cc:88:47: error: 'boolean_t' was not declared in this scope
   CONVERT_OR_THROW_AND_RETURN(value, enabled, boolean_t);
```

src/mediastreamtrack.cc
taylorhoward92: I've made the change - test failing on some builds. I'll have a look into it over the coming days when I get some time.

markandrus: Thanks, @taylorhoward92. I took a look at the failures, and they're all

```
#
# Fatal error in: ../../download/src/pc/rtpreceiver.cc, line 248
# last system error: 0
# Check failed: !stopped_
# Aborted
```

Line 248 is the first line of the private method `webrtc::AudioRtpReceiver::Reconfigure()`, which is called from `webrtc::AudioRtpReceiver::OnChanged()` and `webrtc::AudioRtpReceiver::SetupMediaChannel(uint32_t ssrc)`. I think, although I haven't confirmed, that the `webrtc::AudioRtpReceiver::OnChanged()` is getting called in response to `webrtc::MediaStreamTrack::set_enabled(bool enable)`.

test/mediastream.js
```diff
@@ -159,6 +159,25 @@ tape('.removeTrack and .addTrack on remote MediaStream', function(t) {
...
+    t.ok(tracks.every(function(track) {
+      return !track.enabled;
+    }), 'enabled is false on all MediaStreamTracks');
+    tracks.forEach(function(track) {
+      track.enabled = true;
+    });
+    t.ok(tracks.every(function(track) {
+      return track.enabled;
+    }), 'all MediaStreamTracks are enabled again');
+    t.end();
```
markandrus: @taylorhoward92 I forgot something in my initial review: we should stop each MediaStreamTrack; otherwise, the Node process may hang.

src/mediastreamtrack.cc
taylorhoward92: @markandrus I'm struggling to find what the issue is that's resulting in a stopped track when changing the enabled value (probably due to my minimal C++ experience). Is there anything that you suspect might be causing it?

markandrus:
@taylorhoward92 I’m not sure the stopped issue, but if you don’t mind me pushing commits to your branch I could try pulling it down and seeing if I can get anywhere.

taylorhoward92:
@markandrus I've just pushed the latest changes. Happy for you to commit to my branch - if you're able to take a look that would be great. Thanks

markandrus:
Here's the backtrace I get:

```
* thread #8, name = 'Thread 0x0x10231af30', stop reason = signal SIGABRT
  * frame #0: 0x00007fff6535c23e libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff65412c1c libsystem_pthread.dylib`pthread_kill + 285
    frame #2: 0x00007fff652c51c9 libsystem_c.dylib`abort + 127
    frame #3: 0x0000000103eb3f49 wrtc.node`rtc::webrtc_checks_impl::FatalLog(file="../../download/src/pc/rtpreceiver.cc", line=248, message="!stopped_", fmt=0x00000001051308c0) at checks.cc:155
    frame #4: 0x0000000103c1f1e2 wrtc.node`webrtc::AudioRtpReceiver::Reconfigure() [inlined] void rtc::webrtc_checks_impl::LogStreamer<>::Call<>(file="../../download/src/pc/rtpreceiver.cc", line=248, message="!stopped_") at checks.h:197
    frame #5: 0x0000000103c1f1b9 wrtc.node`webrtc::AudioRtpReceiver::Reconfigure() [inlined] void rtc::webrtc_checks_impl::FatalLogCall<false>::operator&<>(this=0x0000700002d4a660, streamer=0x0000700002d4a658) at checks.h:272
    frame #6: 0x0000000103c1f19b wrtc.node`webrtc::AudioRtpReceiver::Reconfigure(this=0x00000001022028f0) at rtpreceiver.cc:248
    frame #7: 0x0000000103c1f0f1 wrtc.node`webrtc::AudioRtpReceiver::OnChanged(this=0x00000001022028f0) at rtpreceiver.cc:97
    frame #8: 0x0000000103a5780e wrtc.node`webrtc::Notifier<webrtc::AudioTrackInterface>::FireOnChanged(this=0x0000000102203430) at notifier.h:50
    frame #9: 0x0000000103a56934 wrtc.node`webrtc::MediaStreamTrack<webrtc::AudioTrackInterface>::set_enabled(this=0x0000000102203430, enable=false) at mediastreamtrack.h:37
    frame #10: 0x0000000103bab1c0 wrtc.node`void webrtc::ReturnType<bool>::Invoke<webrtc::AudioTrackInterface, bool (this=0x00007ffeefbf5d70, c=0x0000000102203430, m=39 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00, a1=false)(bool), bool>(webrtc::AudioTrackInterface*, bool (webrtc::AudioTrackInterface::*)(bool), bool) at proxy.h:73
    frame #11: 0x0000000103bab106 wrtc.node`webrtc::MethodCall1<webrtc::AudioTrackInterface, bool, bool>::OnMessage(this=0x00007ffeefbf5d20, (null)=0x0000000000000000) at proxy.h:202
    frame #12: 0x0000000103f20199 wrtc.node`webrtc::internal::SynchronousMethodCall::OnMessage(this=0x00007ffeefbf5c98, (null)=0x0000700002d4adf8) at proxy.cc:33
    frame #13: 0x0000000103f6de30 wrtc.node`rtc::MessageQueue::Dispatch(this=0x000000010231af30, pmsg=0x0000700002d4adf8) at messagequeue.cc:510
    frame #14: 0x000000010419671a wrtc.node`rtc::Thread::ProcessMessages(this=0x000000010231af30, cmsLoop=-1) at thread_darwin.mm:75
    frame #15: 0x0000000104001c3a wrtc.node`CRYPTO_set_locking_callback(func=(0x000000010231af30)) - 18446744069347468229
    frame #16: 0x00000001041965ef wrtc.node`rtc::Thread::PreRun(pv=0x000000010231a1a0) at thread_darwin.mm:58
    frame #17: 0x00007fff65410305 libsystem_pthread.dylib`_pthread_body + 126
    frame #18: 0x00007fff6541326f libsystem_pthread.dylib`_pthread_start + 70
    frame #19: 0x00007fff6540f415 libsystem_pthread.dylib`thread_start + 13
```

markandrus:
An interesting thing is that, when I test locally, `self->_track->state()` returns `kEnded`. So I guess webrtc.org code does not want us to call `set_enabled` on `stopped_`/`kEnded` MediaStreamTracks. The reason the MediaStreamTrack in the test is stopped is because it is a remote MediaStreamTrack returned by the helper function `getMediaStream()` which uses an RTCPeerConnection to create the remote MediaStreamTracks it returns (the RTCPeerConnection closes before `getMediaStream()` returns). In fact, we need to test both the local and remote cases.

markandrus:
According to [mediacapture-main](https://w3c.github.io/mediacapture-main/#dom-mediastreamtrack-enabled):

> Thus, after a MediaStreamTrack has ended, its enabled attribute still changes value when set; it just doesn't do anything with that new value.

Also, in reading about "ending" a MediaStreamTrack, via `stop()` or otherwise, it appears `enabled` is unchanged. I will update tests accordingly. We may even have some web-platform-tests we can enabled.

markandrus:
@taylorhoward92 the problem was that we can't call `set_enabled` on stopped, remote MediaStreamTracks. So what I did was this

1. Added a boolean, `_enabled`, to `node_webrtc::MediaStreamTrack`.
2. This boolean is only used for stopped MediaStreamTracks. If `enabled` is get or set on a stopped MediaStreamTrack, we access `_enabled` instead of the MediaStreamTrack.
3. Allow overriding `Stop` so that `node_webrtc::MediaStreamTrack` can set `_ended` and `_enabled` when stopping (this also allowed cleaning up some code).
4. Refactored the tests to test local and remote MediaStreamTracks (since only the remote MediaStreamTracks were demonstrating this issue).
5. Updated CHANGELOG.md.

Can you please review? Then, if there are no further changes and CI passes, we can squash and merge.

taylorhoward92 (commented):
@markandrus thanks for getting to the bottom of that. I was confused because I also read that enabled can still be set on an 'ended' MediaStreamTrack. Anyway, I think your workaround is logical.

Issue #478, opened 2019-01-19
Programmatic Audio


markandrus:
Support for Programmatic Video is coming along [here](https://github.com/node-webrtc/node-webrtc/pull/472). We will want support for Programmatic Audio, too, I think. In the same way that the Programmatic Video PR provides `RTCVideoSource` and `RTCVideoSink` for interacting with "raw"/I420 video frames, I think we want `RTCAudioSource` and `RTCAudioSink` for interacting with raw audio samples. Open question: can we expect to be able to support synchronization?

markandrus:
Implemented in [v0.3.6](https://github.com/node-webrtc/node-webrtc/releases/tag/v0.3.6).

PR #479, opened 2019-01-28
Programmatic Audio

markandrus:
Work in progress. Will address #478.

**Update 1:** Next step is to hook up RTCAudioSource, following [these instructions](https://groups.google.com/d/msg/discuss-webrtc/f3lIyA3Otyo/CmntzcgwAgAJ).

**Update 2:** I can push audio data to the RTCAudioSource and receive it on the RTCAudioSink, locally, now (I haven't tested remote). The test for this is checked in; however, it looks like a pointer is getting double-freed. I don't know the reason yet.

**Update 3:** Double-free issue is resolved. There are a few things I want to do next

- [x] ~Create a spectrogram example (useful for interactively and visually validating the behavior of RTCAudioSink)~ **Update:** I’m just going to do a simple “frequency” detector for simple, clean input sources like a sine wave. It will pair well with the RTCAudioSource example described below. I have a version working, but it’s returning invalid results.
- [x] ~Create an RTCAudioSource example. This should synthesize or otherwise modify audio. It could be a delay effect or, more simply, an oscillator.~ I took a break from this PR to refactor the few example apps we currently have. I’ll soon update this PR to include that. The new example apps include an RTCAudioSource-based sine wave.
- [x] ~Review, refine, and document the APIs for RTCAudioSource and RTCAudioSink.~ See Update 5 for recent notes.

**Update 4:** I've checked in the example application I wrote to help validate the RTCAudioSource/RTCAudioSink APIs. The example application and accompanying tests revealed that the RTCAudioSource/RTCAudioSink APIs ought to be using Int16Arrays rather than Uint8Arrays. Also, the example application should demonstrate to others how to use these APIs to do interesting things. I'll return to work on "Review, refine, and document the APIs for RTCAudioSource and RTCAudioSink." from Update 3.

Screen Shot 2019-03-16 at 5 58 19 PM

**Update 5:** I wanted to confirm stereo behavior where `channelCount === 2` before refining the API, so I added a sine-wave-stereo example:

Screen Shot 2019-03-17 at 2 34 21 PM

Some things I've learned about the audio APIs:

* If a MediaStreamTrack is not being sent to an RTCPeerConnection, its source can push audio data with a variety of `bitsPerSample` and `numberOfFrames`.
* However, once attached to an RTCPeerConnection, the webrtc.org code expects 10 ms of audio data and and `bitsPerSample` to always be 16-bit.
* If we accept this restriction for all sources (because these APIs aren't particularly useful otherwise), and we let `numberOfChannels` default to 1 (seems reasonable), then we can simplify the RTCAudioData dictionary as follows:

  ```webidl
  dictionary RTCAudioData {
    required Int16Array audioData;
    required unsigned short sampleRate;
    octet bitsPerSample = 16;
    octet numberOfChannels = 1;
    unsigned short numberOfFrames;
  };
  ``` 

* If there's a default `sampleRate` that makes sense (I'm not sure there is) we could further simplify and only require the user to provide `audioData`.

* [x] Should `audioData` be renamed `samples`?
* [x] ~Should `numberOfFrames` be renamed `numberOfSamples` or otherwise removed? (It should be redundant with `audioData.length`.)~ Keeping `numberOfFrames` as-is but marking it optional. If `numberOfSamples` was a thing, it'd really be `channelCount * numberOfFrames`.
* [x] ~Should `bitsPerSample` be removed? (It is kind of redundant what with each element of Int16Array being 16-bit&hellip;)~ Keeping `bitsPerSample` as-is but marking it optional.
* [x] Should `numberOfChannels` be renamed `channelCount` to be more consistent with the Web Audio API?

* If the above three changes were made, we'd have the following:

  ```webidl
  dictionary RTCAudioData {
    required Int16Array samples;
    required unsigned short sampleRate;
    octet bitsPerSample = 16;
    octet channelCount = 1;
    unsigned short numberOfFrames;
  };
  ```


src/converters/dictionaries.cc
```diff
@@ -695,6 +695,53 @@ CONVERTER_IMPL(node_webrtc::I420ImageData, rtc::scoped_refptr<webrtc::I420Buffer
   return node_webrtc::Pure(CreateI420Buffer(value));
 }
 
+static node_webrtc::Validation<node_webrtc::RTCOnDataEventDict> CreateRTCOnDataEventDict(
+    v8::ArrayBuffer::Contents audioData,
+    int bitsPerSample,
+    int sampleRate,
+    int numberOfChannels,
+    int numberOfFrames
```
markandrus: These should really be `size_t`.

src/converters/dictionaries.cc
```diff
@@ -760,3 +807,21 @@ TO_JS_IMPL(webrtc::VideoFrame, value) {
...
 }
+
+TO_JS_IMPL(node_webrtc::RTCOnDataEventDict, dict) {
+  Nan::EscapableHandleScope scope;
+
+  auto isolate = Nan::GetCurrentContext()->GetIsolate();
+  auto byteLength = dict.numberOfChannels * dict.numberOfFrames * dict.bitsPerSample / 8;
+  auto arrayBuffer = v8::ArrayBuffer::New(isolate, dict.audioData.get(), byteLength, v8::ArrayBufferCreationMode::kInternalized);
+  auto uint8Array = v8::Uint8ClampedArray::New(arrayBuffer, 0, byteLength);
+  dict.audioData.reset();
```
markandrus: So, converting an `node_webrtc::RTCOnDataEventDict` to a `v8::Local<v8::Value>` will take ownership of the underlying `std::shared_ptr<uint8_t>` and set it to `nullptr` on the `node_webrtc::RTCOnDataEventDict`.

src/converters/dictionaries.cc
```diff
@@ -760,3 +807,21 @@ TO_JS_IMPL(webrtc::VideoFrame, value) {
...
+
+  auto object = Nan::New<v8::Object>();
+  object->Set(Nan::New("audioData").ToLocalChecked(), uint8Array);
+  object->Set(Nan::New("bitsPerSample").ToLocalChecked(), node_webrtc::From<v8::Local<v8::Value>>(dict.bitsPerSample).UnsafeFromValid());
+  object->Set(Nan::New("sampleRate").ToLocalChecked(), node_webrtc::From<v8::Local<v8::Value>>(dict.sampleRate).UnsafeFromValid());
+  object->Set(Nan::New("numberOfChannels").ToLocalChecked(), node_webrtc::From<v8::Local<v8::Value>>(static_cast<double>(dict.numberOfChannels)).UnsafeFromValid());
+  object->Set(Nan::New("numberOfFrames").ToLocalChecked(), node_webrtc::From<v8::Local<v8::Value>>(static_cast<double>(dict.numberOfFrames)).UnsafeFromValid());
```
markandrus: We should add support for `size_t`.

src/rtcaudiosink.cc
```diff
@@ -0,0 +1,105 @@
...
+    std::move(audio_data_copy),
+    bits_per_sample,
+    sample_rate,
+    number_of_channels,
+    number_of_frames
+  }));
+}
+
+void node_webrtc::RTCAudioSink::HandleOnDataEvent(const node_webrtc::OnDataEvent& event) {
+  auto maybeValue = node_webrtc::From<v8::Local<v8::Value>>(event.dict);
```
markandrus, line 87: This is missing a `Nan::HandleScope`.

src/converters/dictionaries.cc
markandrus: Turns out `reset` doesn't do what I want.

markandrus (commented):
Notes to self.

vpalmisano:
Hi Mark,
I'm using this branch to test RTCAudioSource. It is not clear if audioData should be an Unsigned 8 bit Int Array (Uint8ClampedArray as I see in the unit test). In the case of 16 bit audio frames, what is the right format? Unsigned 16 bit big or little endian?
Thanks

vpalmisano:
Resolved: I found that audio_coding_module.cc requires that all input audio chunks must have a duration equals to 10ms.

scaret:
Starting about 100 audio source on different peerConnection will cause crash:

```
#
# Fatal error in: ../../download/src/audio/audio_send_stream.cc, line 330
# last system error: 1
# Check failed: !race_checker.RaceDetected()
# Aborted (core dumped)
```

markandrus:
@scaret I’ve seen that, too, but with just 1 source. Not sure what it is yet

markandrus:
@vpalmisano I’ve been busy but returned to this today. As you mention, you need 10 ms worth of samples, I think you need a sample rate of 48000, and bits per sample should be 16. AFAIK, endianness is same as host, though I’ve only tested on little endian. If I’m wrong, please let me know or share your findings. I’m currently working to just write an RTCAudioSource based sine wave generator to validate behavior.

markandrus:
Sample rate can be 8000, 16000, 32000, or 48000. There's an issue in the C++ code where we return/accept Uint8Array instead of an Int16Array (the samples are signed, not unsigned). I'll fix this week.

vpalmisano:
I'm using a source at 48000 Hz, 16 bit little endian with no issues. At the receiver side, I've successfully tested Google Chrome on Linux and Safari on MacOS and iOS.

markandrus:
Thanks for testing, @vpalmisano. With the exception of the problem @scaret identified (which I still need to look into) and some minor details, I think this is getting close to being ready to merge. Before doing so, I wanted to propose an API simplification, which is in the PR description above under Update 5. Basically, the RTCAudioData dictionary would change to the following:

```webidl
dictionary RTCAudioData {
  required Int16Array samples;
  required unsigned short sampleRate;
  octet bitsPerSample = 16;
  octet channelCount = 1;
  unsigned short numberOfFrames;
};
```

I wonder if you or anyone else watching has any preference/concerns?

**Edit:** I went ahead and implemented the change in a commit. It means in some cases, you only have to do the following to use RTCAudioSource:

```js
const source = new RTCAudioSource();

const sampleRate = 8000;
const samples = new Int16Array(sampleRate / 100);  // 10 ms of 16-bit mono audio

source.onData({ samples, sampleRate });
```

markandrus:
I've pushed a fix for the crash identified by @scaret, along with an integration test in the example app that triggered that particular crash (it's a "stress test" that creates 100 RTCAudioSources). Since I got a thumbs up from @vpalmisano on the API change, assuming all the builds pass, I'll merge this and prepare to release the next version of node-webrtc.

t-mullen:
I'm trying to dump this to FFMPEG/VLC/etc but it's just noise (video works great!).

To be clear, the audio samples are signed 16-bit little-endian PCM audio, right?

pradeep538:
Could you please help me how to access Camera,Mic and speaker for remote audio by example

PR #488, opened 2019-02-21
Allow node to access the VideoFrame rotation

paul-maxime:
The `frame` object on the node.js side, retrieved from `RTCVideoSink`, contains the following elements:
- width
- height
- data

This commit adds a 4th element `rotation`, to get the video orientation from the node.js code.


markandrus:
Thanks @paul-maxime! So the values for `rotation` can be `0`, `90`, `180`, or `270`, per the following?

```cpp
// enum for clockwise rotation.
enum VideoRotation {
  kVideoRotation_0 = 0,
  kVideoRotation_90 = 90,
  kVideoRotation_180 = 180,
  kVideoRotation_270 = 270
};
```

paul-maxime:
Yes, that is correct.

PR #513, opened 2019-04-25
Set up CI with Azure Pipelines

markandrus:


markandrus:
Azure has a 10 GB disk limit which is insufficient for checking out and building webrtc.org code. Closing.

PR #536, opened 2019-05-29
Image stride fix

### LINKED ISSUE #514: i420ToRgba fails when sending 540p using H264
There is a `byteLength` check in `i420_helpers.cc`. So far in my testing, this is working well for frames with a resolution of `1280x720` and `640x360`. When using a resolution of `960x540` like below, this check fails with the following error:

```
Expected a .byteLength of 777600, not 829440
```

```
await navigator.mediaDevices.getUserMedia({
  video: { width: { exact: 960 }, height: { exact: 540 } }
});
```

Are we certain that the following is always true? It appears that for `960x540`, the frame bytes are coming in with a multiple of `1.6` instead of `1.5`.

```
auto expectedByteLength = static_cast<size_t>(imageData.width * imageData.height * 1.5);
```

anshul-kai:
This issue shows up when using H264 at 540p. Fixes #514.

### DIFF
--- a/src/dictionaries/webrtc/video_frame_buffer.cc
+++ b/src/dictionaries/webrtc/video_frame_buffer.cc
@@ -34,11 +34,14 @@ TO_NAPI_IMPL(rtc::scoped_refptr<webrtc::I420BufferInterface>, pair) {
   Napi::EscapableHandleScope scope(env);
   auto value = pair.second;
 
-  auto sizeOfYPlane = value->StrideY() * value->height();
-  auto sizeOfUPlane = value->StrideU() * value->height() / 2;
-  auto sizeOfVPlane = value->StrideV() * value->height() / 2;
+  auto sizeOfSrcYPlane = value->StrideY() * value->height();
+  auto sizeOfSrcUPlane = value->StrideU() * value->height() / 2;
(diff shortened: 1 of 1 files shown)


markandrus:
This looks OK to me. I think we could be more efficient if we copied `stride` worth of bytes rather than looping and copying only `width` worth. That would require the user to deal with stride. This is fine for now, though. I’ll try to release this weekend. Thank you.

anshul-kai:
@markandrus Any reason why this is blocked from being merged?

markandrus:
@a-koka I've just released this! Thanks again! Only reason it lingered there was because I've been very busy/don't have as much time for this project. I'm interested in bringing on someone to help maintain it.

anshul-kai:
Thanks @markandrus. I wish I had the skillset and the time needed to take over this repo. I will definitely help when and where I can.

tdrz:
Thank you for the fix. Any idea when will this be released on npm?

markandrus:
It’s released

tdrz:
I just noticed that actually this only fixes the stride problem. h264 still needs to be enabled manually, right?

anshul-kai:
Yes, h264 still needs to be enabled.

PR #548, opened 2019-08-23
Pass stream ID arrays when calling peerConnection.addTrack and addTransceiver

### LINKED ISSUE #546: Application-specific Track Names
Our application requires the ability to provide custom names to WebRTC tracks that are readable/accessible to the other side of the session/call.

The underlying WebRTC `PeerConnectionInterface::AddTrack()` method allows the caller to pass in a vector of strings representing stream labels to associate the track with.  This allows the remote side to retrieve the stream the track belongs to and to read the steam name.  However, the current node-webrtc `RTCPeerConnection::AddTrack()` method expects an optional `MediaStream` object pointer, from which a string identifier is extracted.  Node-webrtc users have no access to `MediaStream` objects themselves, so this optional parameter appears to serve no purpose today. _(Edit: I now see that we can create MediaStream objects from node-webrtc and use them, but the `id` field is auto-generated and read-only.)_

We have modified `RTCPeerConnection::AddTrack()` in rtc_peer_connection.cc to allow passing in a string instead of a MediaStream object pointer, to more closely mirror the underlying API (see "// NEW:" comments in snippet below).  Ideally this would be a vector of strings to model the native API's, but we would like feedback on this approach, before submitting a GIT pull request.

```c++
Napi::Value RTCPeerConnection::AddTrack(const Napi::CallbackInfo& info) {
  auto env = info.Env();
  if (!_jinglePeerConnection) {
    Napi::Error(env, ErrorFactory::CreateInvalidStateError(env, "Cannot addTrack; RTCPeerConnection is closed")).ThrowAsJavaScriptException();
    return env.Undefined();
  }
  // NEW: third arg includes a maybe string instead of MediaStream
  CONVERT_ARGS_OR_THROW_AND_RETURN_NAPI(info, pair, std::tuple<MediaStreamTrack* COMMA Maybe<std::string>>)
  auto mediaStreamTrack = std::get<0>(pair);
  // NEW: extract the string instead of MediaStream
  Maybe<std::string> mediaStreamID = std::get<1>(pair);
  std::vector<std::string> streams;

  if (mediaStreamID.IsJust()) {
    // NEW: add the new string / ID directly to the streams vector
    streams.push_back(mediaStreamID.UnsafeFromJust());
  }
  auto result = _jinglePeerConnection->AddTrack(mediaStreamTrack->track(), streams);
  if (!result.ok()) {
    CONVERT_OR_THROW_AND_RETURN_NAPI(env, &result.error(), error, Napi::Value)
    Napi::Error(env, error).ThrowAsJavaScriptException();
    return env.Undefined();
  }
  auto rtpSender = result.value();
  return RTCRtpSender::wrap()->GetOrCreate(_factory, rtpSender)->Value();
}
```

Please let us know what you think of this approach.  Thank you.

csheely:
closes #546 

This pull request contains our proposal to supply media stream IDs from the javascript peer connection addTrack() and addTransceiver() methods.  IDs are passed in as an array, which is similar to the underlying WebRTC API.


lib/peerconnection.js
```diff
@@ -204,8 +204,8 @@ RTCPeerConnection.prototype.addTransceiver = function addTransceiver() {
   return this._pc.addTransceiver.apply(this._pc, arguments);
 };
 
-RTCPeerConnection.prototype.addTrack = function addTrack(track, stream) {
-  return this._pc.addTrack(track, stream);
+RTCPeerConnection.prototype.addTrack = function addTrack(track, streams) {
+  return this._pc.addTrack(track, streams);
```
markandrus, line 208: Please check out the Web IDL for `addTrack` [here](https://w3c.github.io/webrtc-pc/#rtcpeerconnection-interface-extensions) — `streams` should use the [rest parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) syntax. So, at least in the JavaScript layer here, we should define this method as

```js
RTCPeerConnection.prototype.addTrack = function addTrack(track, ...streams) {
  return this._pc.addTrack(track, streams);
};
```

node-webrtc supports Node 8 onwards, so I believe we ca safely use rest parameter syntax here.

src/dictionaries/node_webrtc/rtc_media_stream_init.cc
```diff
@@ -0,0 +1,18 @@
...
+static Validation<RTC_MEDIA_STREAM_INIT> RTC_MEDIA_STREAM_INIT_FN(
+    const std::string& id) {
+  return Pure<RTC_MEDIA_STREAM_INIT>({id});
+}
+
+}  // namespace node_webrtc
+
+#define DICT(X) RTC_MEDIA_STREAM_INIT ## X
+#include "src/dictionaries/macros/impls.h"
+#undef DICT
```
markandrus, line 18: 👍

src/dictionaries/node_webrtc/rtc_media_stream_init.h
```diff
@@ -0,0 +1,16 @@
...
+// IWYU pragma: no_include "src/dictionaries/macros/impls.h"
+
+#define RTC_MEDIA_STREAM_INIT RTCMediaStreamInit
+#define RTC_MEDIA_STREAM_INIT_LIST \
+  DICT_REQUIRED(std::string, id, "id")
+
+#define DICT(X) RTC_MEDIA_STREAM_INIT ## X
+#include "src/dictionaries/macros/def.h"
+#include "src/dictionaries/macros/decls.h"
+#undef DICT
```
markandrus, line 16: 👍

test/destructor/index.js
```diff
@@ -109,9 +109,9 @@ test('Destructors fire in MediaStreamTrack use-case', async t => {
 
     const [pc1, pc2] = await negotiateRTCPeerConnections({
       withPc1(pc1) {
-        stream1.getTracks().forEach(track => pc1.addTrack(track, stream1));
+        stream1.getTracks().forEach(track => pc1.addTrack(track, [stream1]));
```
markandrus: Per comment above, using rest parameter syntax, there should be no change necessary here.

test/mediastream.js
```diff
@@ -75,6 +75,13 @@ tape('new MediaStream(tracks)', function(t) {
   });
 });
 
+tape('new MediaStream(mediaStreamInit)', function(t) {
+  var stream = new MediaStream({ id: 'testStreamId' });
```
markandrus, line 79: Ca you please include a test to see what happens when two MediaStreams with the same ID are created? Also, I'm curious what happens if two MediaStreams with the same ID are used on an RTCPeerConnection.

test/rtcrtpsender.js
```diff
@@ -43,45 +42,45 @@ var sdp = [
   'a=ssrc:2 msid:stream 456'
 ].join('\r\n') + '\r\n';
 
-tape('.addTrack(track, stream)', function(t) {
+tape('.addTrack(track, [stream])', function(t) {
```
markandrus: Per comment above, using rest parameter syntax, there should be no change necessary here. Can you please revert all such changes below?

markandrus (commented):
Overall looks good. Thanks for y'all's help here. There is one set of changes we still need related to rest parameter syntax in the JavaScript layer. Once that is completed, I believe we can merge.

lib/peerconnection.js
sgodin: Will do.

test/mediastream.js
sgodin: Sorry I should have included some details on this.  I did set out to do this, however I backed it out.  Here's what I did:
-I analyzed the webrtc.org code base for what would happen if 2 MediaStreams had the same id.  I discovered that there was no major concern here.  It appears that worse case the 2nd stream would fail to get added to a peer connection if one with the same name already existed.
-I thought about adding a test case to test/mediastream.js - however this file really just deals with MediaStream objects in isolation.  So even though I could create test case where we create 2 media streams with the same id, we wouldn't really be testing anything useful, until they are added to a track or transceiver.
-I then set out to add such test to test/rtcrtpsender.js (since it has addTrack tests in it), however I was unable to come up with any meaningful way to verify anything (ie: assert conditions), since I can't actually retrieve the stream id, from an interface outside the MediaStream object itself, once it is set.  Also - we know that addTrack and addTransceiver in the C++ side don't really do anything with the media stream object, other than extracting the id from it.

If you have some other idea's or approaches I could use, I would be happy to try to implement them.  :)

Scott

sgodin: OK, I found a way to verify things by looking at the generated SDP.  I added a new test case to rtcrtpsender.js - let me know what you think.  :)

markandrus (approved):
Thanks so much!!!

PR #566, opened 2019-10-22
Video timestamp fix

lonocvb:
`VideoFrame` is built with default timestamp, 0 in practice, it results the  incorrect timestamp is reported in RTCP packages.

The relevant code in `libwebrtc`//modules/rtp_rtcp/source/rtcp_sender.cc:456
```c++
  // Round now_us_ to the closest millisecond, because Ntp time is rounded
  // when converted to milliseconds,
  uint32_t rtp_timestamp =
      timestamp_offset_ + last_rtp_timestamp_ +
      ((ctx.now_us_ + 500) / 1000 - last_frame_capture_time_ms_) * rtp_rate;
```
Here the `last_frame_capture_time_ms_` is updated by the latest `VideoFrame`. It is 0 and results an incorrect `rtp_timestamp` in RTCP packages obviously.

### DIFF
--- a/src/interfaces/rtc_video_source.cc
+++ b/src/interfaces/rtc_video_source.cc
@@ -20,6 +20,9 @@
 #include "src/functional/maybe.h"
 #include "src/interfaces/media_stream_track.h"
 
+#include <chrono>
+#include <ctime>
+
 namespace node_webrtc {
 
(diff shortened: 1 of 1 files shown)


markandrus:
Thanks!

PR #634, opened 2020-05-24
fix: add Node 14 into engines

farnabaz:
wrtc already supports Node-14 as it passes all the tests, but because of `engines` in `package.json` using it on Node-14 is restricted


markandrus:
Ah crud, I missed this. Thank you!

PR #644, opened 2020-07-07
iceevent srcElement property

CharlesRA:
Hello.

I added the srcElement object in the event's data. This property can be found in the original version of RTCPeerConnectionIceEvent

### DIFF
--- a/lib/peerconnection.js
+++ b/lib/peerconnection.js
@@ -18,24 +18,6 @@ function RTCPeerConnection() {
 
   EventTarget.call(this);
 
-  this.getSrcElement = function() {
-    return {
-        ...this,
-        canTrickleIceCandidates: this.canTrickleIceCandidates,
-        connectionState: this.connectionState,
-        currentLocalDescription: this.currentLocalDescription,
-        currentRemoteDescription: this.currentRemoteDescription,
-        iceConnectionState: this.iceConnectionState,
-        iceGatheringState: this.iceGatheringState,
-        localDescription: this.localDescription,
-        pendingLocalDescription: this.pendingLocalDescription,
-        pendingRemoteDescription: this.pendingRemoteDescription,
-        remoteDescription: this.remoteDescription,
-        sctp: this.sctp,
-        signalingState: this.signalingState
-    };
-  };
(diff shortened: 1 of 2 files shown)


lib/peerconnection.js
```diff
@@ -49,15 +67,15 @@ function RTCPeerConnection() {
   };
 
   pc.onsignalingstatechange = function onsignalingstatechange() {
-    self.dispatchEvent({ type: 'signalingstatechange' });
+    self.dispatchEvent({ type: 'signalingstatechange', srcElement: self.getSrcElement() });
```
markandrus: Hmm, why wouldn't we just pass `this` as `srcElement`?

lib/peerconnection.js
```diff
@@ -18,6 +18,24 @@ function RTCPeerConnection() {
 
   EventTarget.call(this);
 
+  this.getSrcElement = function() {
```
markandrus: Do we really need this? It's not a standard API, so I prefer not to add it.

markandrus (commented):
Thanks for the contribution. I have some concerns about the implementation. Also, why is `srcElement` useful? Checking [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Event/srcElement), I see this is no longer recommended. The standard property should be `target`, and so I think we should have support for that first. Moreover, `srcElement` should probably be implemented on some parent Event class that RTCPeerConnectionIceEvent extends as a property that derives from `target`. WDYT?

CharlesRA:
Thank you for your reply. I will replace srcElement with target.

lib/peerconnection.js
CharlesRA: I have made the changement

lib/peerconnection.js
CharlesRA: Removed.

markandrus:
Thank you!

CharlesRA:
Thank you for the approve, any idea of the V.0.4.6 release date ?

markandrus:
I'm working on it ATM. It should be out today or tomorrow.

PR #656, opened 2020-10-04
Add typescript type definitions

3Dcube:


markandrus:
Thanks, @3Dcube!

Aren’t we missing some declarations, though? For example, RTCPeerConnection’s standard `addTrack`, `addTransceiver`, etc.? I don’t see this in the types file.

And any chance we could inherit/extend from existing WebRTC type declarations?

3Dcube:
I'm already using existing types, but maybe I need to reference the standard library explicitly

image

There could be confusion since variable names and type names are the same, but they are different.

```typescript
export declare var RTCPeerConnection: {
    prototype: RTCPeerConnection;
    new(configuration?: ExtendedRTCConfiguration): RTCPeerConnection;
    generateCertificate(keygenAlgorithm: AlgorithmIdentifier): Promise<RTCCertificate>;
    getDefaultIceServers(): RTCIceServer[];
};
```

I'm not declaring type `RTCPeerConnection` here. I'm declaring variable `RTCPeerConnection` which has `new(configuration?: ExtendedRTCConfiguration): RTCPeerConnection;` constructor and returns already existing `RTCPeerConnection` type from the standard library https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L12186

pnutmath (approved):

anteeek:
I'd love to see this merged, is any help needed?

PR #675, opened 2021-01-07
Use range to define `node` engine

piranna:
Instead of adding new versions in `package.json` file `engines.node` field, use a range starting with Node.js 10.x, since we are targeting n-api. This change makes `node-webrtc` automatically compatible with any future Node.js version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "host": "https://node-webrtc.s3.amazonaws.com"
   },
   "engines": {
-    "node": "^8.11.2 || ^10.0.0 || ^11.0.0 || ^12.0.0 || ^13.0.0 || ^14.0.0 || ^15.0.0"
+    "node": "^8.11.2 || >=10.0.0"
   },
   "dependencies": {
     "nan": "^2.3.2",


markandrus:
@piranna I landed this on the support-v0.4.7 branch (there are some unreleasable changes queued up on the develop branch at the moment). After I did so, I noticed an incompatibility with NPM 7. I'm going to leave the release as-is, since NPM <7 is unaffected; but installing `wrtc` with NPM 7 will fail for the moment.

The root cause appears to be an issue with NPM 7 install scripts. Previously, install scripts had access to all dependencies declared by a library. In our case, we declare a dependency on node-pre-gyp and we expect to be able to `spawnSync` node-pre-gyp in the install script; however, NPM 7 doesn't appear to have node-pre-gyp downloaded or on the path&hellip;

I don't have time to open/investigate with NPM at the moment. Perhaps you (or another contributor) could follow up on it?

markandrus:
https://github.com/node-webrtc/node-webrtc/commit/38ca026afc4447fcc6c364617c11bc8c4db56435

PR #691, opened 2021-04-06
Ensure it is possible to create a debug build for Windows

sparkpunkd:
I needed to create a debug build for use on Windows platforms and following the build instructions did not result in a successful build. 

This PR adds the `_HAS_ITERATOR_DEBUGGING=0` definition to debug builds so that the node-webrtc binding objects will link against the external dependencies, where the `gn` debug build is configured with `enable_interator_debugging=false`. 

In addition, I've corrected a couple of details and added some helpful notes to the build instructions.


markandrus (approved):
Thanks!

PR #740, opened 2022-11-05
Fix install by removing bundleDependency (v0.4.7 branch)

chfritz:
See https://github.com/node-webrtc/node-webrtc/issues/721#issuecomment-1304397891.


NareshPendamUVenu (commented):
please resolve this for mac with arm64 processors

